PR #1, opened 2015-01-08
Added newline to exception message

MarkRedeman:


rosstuck:
Ah, nice catch! Maybe a space would make more sense here instead of a new line. That way the logs can wrap it however they see fit. Would you be willing to make that change and I'll gladly merge it? :smile:

MarkRedeman:
Sure. :)

rosstuck:
Thanks! :dancers:

PR #6, opened 2015-01-13
Ignore composer.lock

acairns:
Ignore package composer.lock file


rdohms:
Oh no, please don't.

The `composer.lock` should be included. This allows people who contribute to this class to be on the same page as the developer.

Common believe is that this will affect people who use the library, it won't as lock files of dependencies are ignored by composer. So essentially this file will not in any way affect users.

However, lack of a lock file means that the Scrutinizer/Travis builds are now un-bound. Meaning the builds may have weird results due to picking up different versions, which makes it very confusing to debug. Not to mention people contributing may get different versions and hence random unit test failures not related to their contribution.

The only reason to remove a lock file would be to have Travis run checks against "un-tested versions", but this would happen in a random way that is not effective. For this i suggest setting up a travis configuration that runs a specific build with an `composer update` instead of an install, make it "failable". This would be a future-proofing build.

But please, do not remove lock files, not form libraries, not from projects.

rosstuck:
@rdohms is correct here, the composer.lock file should always be checked in with every project using composer.

acairns:
I was under the understanding they weren't ignored also.

Shit, sorry guys! :(

rosstuck:
No worries, it's a really common mistake. :smiley:

Ocramius:
> This allows people who contribute to this class to be on the same page as the developer.

Disagree here: any `composer.json` supported range of versions should be supported, always.

 > However, lack of a lock file means that the Scrutinizer/Travis builds are now un-bound. 

They are supposed to be like that: eager updating in travis-ci discovers failures caused by dependency changes.

This is also **VERY DANGEROUS**, as users of the library (library as a dependency) are now using a set of dependencies (usually "latest and greatest") that is un-tested with this package unless the maintainer keeps updating the dependencies at every commit.

 > which makes it very confusing to debug

This is what we're supposed to actually do: eagerly debug issues as they come up, and collaborate with other library maintainers in handling BC breaks in their code.

 > The only reason to remove a lock file would be to have Travis run checks against "un-tested versions"

There are no "un-tested" versions: latest dependency set should always be the set of deps that are installed.

 > For this i suggest setting up a travis configuration that runs a specific build with an `composer update` instead of an install, make it "failable".

It should always be like that, and it should not be "failable".

 > But please, do not remove lock files, not form libraries, not from projects.

`.gitignore` of libraries MUST include `composer.lock`, whereas deployable packages must have a committed `composer.lock`.

That said, this PR is perfectly legit and should be merged, IMO.

rdohms:
@Ocramius I disagree with you.

Not every contributor to a library has the time to engage in long dialogues and handling BC discussions, sometimes all they can offer is to fix a bug or another.

I do not believe leaving out the `composer.lock` is the right way to ensure all versions are tested, its random, and it should be controlled. 

Advanced developers that have time to invest into packages can always ensure the future scenarios, but it should not force less engaged contributors to deal with errors not related to their own contributions.

Having a specific build instruction that tests "latest" versions of dependencies offers the same results but within a controlled environment. Much more informative and productive then random breaking.

everzet:
@rdohms you're working against assumption here that every change in dependencies will completely destroy this repository to the extent where people will not be able to make it work. Can we simply agree to not write software in the way that:

1. Gets broken miserably when supported dependency upver breaks
2. Does not communicate which dependency is broken via the test suite

If you're really scared of dependencies breaking your lib and your lib is so fragile that any update can destroy it, then committing `composer.lock` is not the right solution. The right solution is to set exact versions in `composer.json`.

acairns:
I agree with @Ocramius 

If a package is only going to support a specific version of a dependency, shouldn't that be set in composer.json?

mbeccati:
@rdohms if I require a library, will its own composer.lock be taken into account or ignored?

WouterSioen:
It will be ignored.

rdohms:
@everzet if we could all do that we would not even need a composer.lock.

I do agree with both of you and i wish we did not have these situations, i'm just dealing with the real world scenarios i have seen and making a simple recommendation to offer structured breakage, not random.

Ocramius:
> Not every contributor to a library has the time to engage in long dialogues and handling BC discussions, sometimes all they can offer is to fix a bug or another.

These are just bugs. If a bug is present, it needs to be solved, but it may be unrelated with what a contributor is working on.
Discovering a bug is why we have a CI and a dev setup.
Do not hurt the actual stability of the project to make things easier for contributors at first install: everyone contributing to this sort of project is perfectly capable of restricting `composer.json` dependencies if a BC break in a dependency is discovered.

 > I do not believe leaving out the `composer.lock` is the right way to ensure all versions are tested, its random, and it should be controlled. 

Let me make you this particular example to illustrate why `composer.lock` is completely useless for a DEV/CI environment (pseudo-composer-definition):

```json
{
    "package1": {"1.0.0": {"require": {"package2": "2.*"}}},
    "package2": {
        "2.0.0": {"require": {"php": "~5.4.0"}},
        "2.1.0": {"require": {"php": "~5.5"}},
    }
}
```

This is because `package2` provides polyfills for a functionality on version `2.0.0`, whereas version `2.1.0` simply skips over that, providing a raw (fast) functionality, maybe based on generators: all perfectly semver compliant, as it's all internal API.

Now give your environment's (PHP 5.5 or more) `composer.lock` to someone that has PHP 5.4 installed and he will just have a nice fatal error.

 > Advanced developers

Holy crap, we're talking basics of PHP development of 2013. Not 2015, not 2014: 2013.

rdohms:
@mbeccati it will be ignored.

mbeccati:
@rdohms than it won't prevent breakages coming from the dependencies anyway for end users. I would agree with you if we were talking about a deployable project.

rdohms:
@ocramius that is a valid point i have to agree on.

I would appreciate you not distort my "advanced developer" statement. I mean developers who actively participate in the development of the package and those who are "casual" and may not know the package in depth to even identify the issue as caused by the new dependency. For these avoiding random breakage is better.

It does indeed boil down to what you said: focusing on making contributor's life easier or having every build get latest versions.

I personally prefer to offer controlled breakage and have a specific build test updates, but I guess both approaches could be considered valid for libraries. I'll think a bit more on this as i recall other reasons.

@mbeccati like i said in the first post: the common belief that this affects endusers is wrong, so i'm not talking about them at all here.

Ocramius:
> I would appreciate you not distort my "advanced developer" statement.

No distortion, really, these are things that should be familiar to a PHP developer working with composer. Additionally, I prefer having one sad contributor more, and one discovered bug more, rather than a hidden bug and a contributor helping in writing broken code ;-)

@rdohms focus on the actual thing that lands on `packagist.org` and how to make it stable, I'd say.
Developing it is secondary to that.

rosstuck:
Walk away from computer for 15 minutes, come back to viral github thread. :grin: 

Those're good points and @Ocramius and @everzet have a lot of experience maintaining large packages, so I'll take their advice and merge this one. For a project with more dependencies than this one, I could see @rdohms's point about a separate controlled build that clues you in.

But Tactician core is really tiny so I'll just merge this one. Thanks to everyone for your contributions, much appreciated! :smile:

GeeH:
I have no valid opinion on this.

philsturgeon:
This is why I blog.

https://philsturgeon.uk/php/2014/11/04/composer-its-almost-always-about-the-lock-file/

fprochazka:
`composer.lock` should not be included in libraries but in projects. I agree with @Ocramius.

PR #10, opened 2015-01-26
Command Middleware

rosstuck:
So, Command Middleware is something we've shelved previously but I want to reopen the issue after some discussions at PHPBenelux. I've also updated the core package with a sample implementation in this PR.

For those of you not familar with it, Command Middleware is a Chain-of-Responsibility version of extension rather than Decorators. @matthiasnoback's SimpleBus was the first to use a CoR approach rather than Decorators, but it was originally mutable which is something I really didn't like. During some private discussions, I suggested a PSR-7 style middleware approach and sketched out how it could work. That ultimately made its way to SimpleBus 2.0 and I'm suggesting it here as well.

There are two main reasons to use this approach:
- The big feature is that we can progmatically attach plugins. There's no more custom factories or bulky decorator chain setup in your DI container, we can provide a single consistent interface for attaching plugins. More below.
- It means all plugins are on equal ground. @sagikazarmark and I had this discussion with the RemoteCommandBus and HandlerCommandBus, about which should be the end of their decorator chains or should they all pass through. Depending on how we solve an open point below, this could be resolved everywhere.

What changed my mind? I implemented a logging and doctrine transaction decorator at PHPBenelux, but when I started assembling them in a simple script, it became extremely bulky. Also, while talking to different users, I realized the current model is....okay-ish...for some imperative DI containers (ZF2, Laravel) but is downright awful for declarative DI containers (like Symfony).

Middleware means you can attach them with a common interface but the next hurdle is making certain the middleware priority is correct. I wanted to avoid priority numbers ala ZF1 at all costs, so the other insight I had was that plugin extension can be done not with tags, but as a master app config that lists which services to load from the DI container. That's simply an array, which is easier to implement, easy for end users and has order. It also means that we're sideloading the dependencies, which might make it possible to have framework integration bits ship with optional, pre-made config for plugin packages (That all might be a bit of a Symfony specific detail but it's what I'm used to and it means that it works in both imperative and declarative DI containers).

The main changes are:

- CommandBus: https://github.com/thephpleague/tactician/blob/8a8df0695fe7816d0e38b52698747c475b064c2c/src/CommandBus.php 
- Middleware: https://github.com/thephpleague/tactician/blob/8a8df0695fe7816d0e38b52698747c475b064c2c/src/Middleware.php
- HandlerMiddleware: https://github.com/thephpleague/tactician/blob/8a8df0695fe7816d0e38b52698747c475b064c2c/src/HandlerMiddleware.php
- LockingMiddleware: https://github.com/thephpleague/tactician/blob/8a8df0695fe7816d0e38b52698747c475b064c2c/src/LockingMiddleware.php

Open points:

- Do we like the current implementation?
- If no middleware breaks the chain, should we throw an "UnhandledException" or simply assume no-op? An exception might clarify bad configuration to users but it means some implementations are "terminal" again.
- Right now, the next closures passed in don't require the Command be passed to them. We could make the Command a parameter (optional or mandatory) on these callables to allow Middleware to change or replace the Command given further down the chain. Should we allow that? It might also allow us to only generate the ExecutionChain once.
- Should we use a "callable" typehint everywhere or are we better served with Closure (which can be relied upon now I think)?
- Samples, docs and other packages need to be updated, I'd be glad to take the lead on it. Porting from decorator to Middleware is pretty trivial.


rosstuck:
ping @acairns @frankdejonge

sagikazarmark:
I would say we shouldn't force a developer to use either middleware or decorator pattern. We can say that our primary preference is a middleware chain, but that should still allow using the decorator pattern. So I would keep the ComandBus interface (which is removed now).

I see potential in the middleware design. However I have to check it carefully. There are some things I don't really understand. For example the next callback thing.

I also don't really understand how PSR7 is related to this. Perhaps I missed another complete-rewrite-from-scratch period.

I think callable typehint is better than Closure.

Not sure about the unhandled case. It might be exceptional, but what if the command is a self-executing one?

A technical matter: shouldn't middlewares be moved under the Middleware namespace?

matthiasnoback:
Excellent work! A few comments:
- Introducing middleware has indeed turned out to be a great move for SimpleBus.
- I have kept the generic interface for the command bus (in my case it's called `MessageBus`), which would allow people to still use decoration if they want.
- I am a bit unhappy that I have introduced middleware service definitions with priorities to be able to load them in the correct order - these priority numbers are arbitrary, which I don't like in my code ;) What you really need is a way to guarantee that middleware A is being called *before* or *after* middleware B. This should be easy to accomplish I think, and I am considering it for the next SimpleBus at least (it may need some trying out in practice).
- The "unhandled" case is interesting. Still, I think that either a) there is a middleware which resolves a command handler, and if it can't find one, throws an exception (which points out your configuration mistake) or b) there is no middleware which resolves a command handler, and in that case there is no canonical way in which you can tell if a command has been handled or not. It may have been shipped off to a message queue or something. So, I decided not to represent this case in the code. The best thing I might have done is catching the "a)" exception and rethrowing it again more explicitly as "CommandNotHandled".

sagikazarmark:
The question is that the priority should be a static dependency tree (A should ALWAYS run befor/after B) or should be specified in all cases. Both have pros/cons.

rosstuck:
Hi guys,

Thanks for the feedback.

- I'm okay with keeping the CommandBus interface, though I'm not sure what we would need it for anymore. Any suggestion on names? MiddlewareCommandBus sounds a little dopey since Middleware should become first-class citizens in Tactician. 
- I went with an exception as the last `$next` because it seemed easiest to the user. We do get terminal middleware again, but I'm not sure how that'd work with return values if we didn't have that anyways.
- I'm in favor of dependencies just being a plain list of middleware that are always executed in that order.

@matthiasnoback Thanks! For the Tactician Symfony bundle, I'm going to suggest that we move the middleware order from the DI config to the app.yml config as a list of service ids, like so:
~~~yml
tactician:
    middleware:
        - tactician.middleware.logging
        - tactician.middleware.locking
        - my_app.custom.middleware
        - tactician.doctrine.middleware.transactions
        - tactician.middleware.handlers
~~~

That's not necessarily idiomatic Symfony style but it solves the order issue perfectly, it's super easy for an end user to tweak and super easy for us to ship configs for. Plus, as I mentioned, we can implement the same style for basically every framework out there.

rosstuck:
@sagikazarmark Sorry, for the PSR-7 remark, I just meant a type of object setup that's been discussed for HTTP middleware before. I'm not sure what to call this exact setup, so I was just throwing it out there as a comparison. There's no middleware interface in PSR-7 (though I dream of one for a later PSR.... :heart: )

rosstuck:
Oh, and I'm in favor of moving them to a middleware namespace.

For self executing commands, we'd have a self-execution middleware that checks for the SelfExecutingCommand interface and triggers the execute method.

sagikazarmark:
interface: CommandBus
class: [CommandBus]Stack?

src/Setup/QuickStart.php
```diff
@@ -29,11 +29,13 @@ class QuickStart
...
-        $handlerCommandBus = new HandlerCommandBus(
+        $handlerMiddleware = new HandlerMiddleware(
             new InMemoryLocator($commandToHandlerMap),
             new HandleInflector()
         );
 
-        return new LockingCommandBus($handlerCommandBus);
+        $lockingMiddleware = new LockingMiddleware();
+
+        return new CommandBus([$handlerMiddleware, $lockingMiddleware]);
```
stivni: Shouldn't you first lock and then handle?

stivni:
Ok guys, after reading the code and this pull request I decided to throw away what I had in my CommandBus development. This is definitely going in the right direction!

Let me give my opinion on the points discussed above:

- I like the middleware/chain implementation.
- I think it's better not to throw the `CommandWasNotHandledException` when the command was not handled. It could be intended. It is indeed better to rely on the HandlerMiddleware to throw if no handler was found. If we really think we should prevent users from making mistakes, we could extract a `GuaranteedExecutionMiddleware` and chain it to the end in the default setup. So it can still be left out if necessary.
- For me it is a must to be able to convert the command before passing it along to the next middleware. Imagine API changes for with middleware could convert command v1 to command v2, then I would like to pass the new command further on. So making command a (mandatory) parameter would be logic to me. Giving that command is no longer part of the closure's context, this enables you to move the `createExecutionChain` to the constructor, which also seems more logical, as this is part of initialization, not of execution.
- I see no problems in typehinting on `Closure` everywhere.
- Maybe it could be usefull to keep the `CommandBus` interface, but if everyone agrees that middleware is the way to go, I see no use for it.
- I like passing the chain with a plain list.
- I my code, the class that is dispatching commands to command handlers is called `CommandDispatcher`, it would like it better as a name for `HandlingMiddleware`. In fact, I don't like the word `Middleware` in it. Maybe it is just because I'm not really used to the term.
- I agreed that the provided middleware should be moved to its own namespace. I even suggest moving it into (a) plugin(s). Probably even in its own project(s), keeping @matthiasnoback's principles of PHP package design in mind.

For me, the API looks more or less stable, so I would suggest to merge once documentation and examples are adapted.

Keep up the good work!

acairns:
This is really great stuff Ross, good work.

Some points:
- I like the implementation and the simple array configuration
- Agree with @matthiasnoback about priorities being arbitrary
- Having middleware to check for SelfExecutingCommand and trigger execute would work nicely
- Exception wise, I agree with @stivni that it would be best not to throw it in favour of having it part of the default stack. Also feel the same as @matthiasnoback with the handler-resolving middleware being responsible for throwing the exception. `HandlerMiddleware` is would do that and works nicely.

I've been back-and-forward on passing the command to the closure during the middleware chain. I'm gonna have to play with each way a little more.

rosstuck:
tl;dr I'm pretty much sold on making the command a required parameter to the closure and removing the default exeption.

@stivni Wow, thanks, that means a lot. :smile: Great catch about the QuickStart order issue! I've corrected that in the PR.

- I like the GuaranteedExecution middleware approach but you're right in pointing out that the HandlerMiddleware (properly attached) does the same thing really. So, maybe there's no point in it.
- The Command conversion is a good use case. I keep leaning towards it, also for the one-time chain setup, so I think we should go that way as well.
- Closure would be a safer typehint since a callable can technically be an array. I wish there was an "invokable" typehint...
- Middleware is most often used as an HTTP-centric term, so we could change it. `Plugin` is so generic as to be useless. There was a long debate about the Handler class name before, and one of the things we'd like to support in Tactician is that Handlers are optional, so it does make sense to have Handler/Handling somewhere in the final classname though.
- Most of the middleware will ship in separate packages and we've started that (like @sagikazarmark's [Bernard plugin](https://github.com/thephpleague/tactician-bernard-queueing)  that offers queueing support. The pure PHP stuff with no dependencies will ship in the core package, but this probably won't grow past what's already in here.

@acairns Sounds good! What would you (or anyone else) think about adding a SelfExecutingCommand interface and middleware to the core package or a plugin package? It's not a style I like per se but there does seem to be some interest out there for it.

sagikazarmark:
@rosstuck this is something similar to the composite CommandBus we talked about earlier. ;)

rosstuck:
Okay, I've removed the fall-through exception and a command is now a required parameter of the next callables. This also means we can build the execution chain once when creating the CommandBus instead of every execution, which saves a little work.

Still to come:
- Move to middleware sub-namespace
- Change typehint `callable` to `Closure`
- CommandBus interface / namechange?

sagikazarmark:
Honestly, I find the next callable thing really weird. Middlewares shouldn't be aware of any other logic than their execution part. It breaks the separation of concerns.

matthiasnoback:
@sagikazarmark The purpose of these middlewares is to allow them to do things right before or after calling the next middleware. They are in control of that (they could even decide *not* to call the next middleware).

sagikazarmark:
Can't it be decided based on an exception thrown or a specific return value? Separation of concern is still not answered. I can accept if we say it worth breaking the separation of concerns, but from that point of view it is definitely a wrong solution.

matthiasnoback:
@sagikazarmark This won't work, because using an exception or a return value means the middleware loses control. Think of something like this:

```php
class Transactional implements CommandBusMiddleware
{
    public function execute(Command $command, Closure $next)
    {
        $this->connection->startTransaction();
        try {
            $next($command);
            $this->connection->commit();
        } catch (Exception $exception) {
            $this->connection->rollback();
        }
    }
}
```

You need absolute control over when the next middleware is going to be called. Using `return` or `throw` wouldn't allow you to do something after the next middleware has been called.

rosstuck:
Indeed. This middleware thing is exactly the same as a decorator: the big advantage is that you can control whether you're running before or after some other operation. 

So, if a decorator was okay, this is equally okay. Now, if we were still saying middleware are decorators of command bus, then the interface would have a dual responsibility issue and I'd agree... but we've split middleware off as a separate concept now so, this is all consistent.

Discussing alternate implementations:

- I personally don't see it as an SRP issue really. The only two responsiblities here would be: a) the proper timing/execution of the sub-operation (calling innerCommandBus or next) and b) whatever the added behavior is (logging, transactions, whatever). But in this case, the two are linked, you can't really separate them out. We could split responsibility B into a separate object, and use it in the middleware as a dependency. I'd still totally recommend doing that if it grows too large so that still applies. **We would absolutely advocate keeping the middlewares light.** Just like event listeners or controllers. It just may not always be worth it for very small things like @matthiasnoback's example above.

- Since this all about the before/after aspect of some other operation, we could change the middleware to have onBefore/onAfterExecution/onFailure methods. But even if that did work, objects would need to store a lot of matching state for some operations, like "was this the same object I saw in the onBefore method now that it's come around to the onAfter method". However, that's really hard and error-prone, especially since a middleware might have subcommands flying through at the same time or be converting them like in @stivni's usecase, etc. With this and the decorator approach, it all stays in one method call so it's pretty much stateless.

That's my argument for this. It is different than what we see in PHP very often, it's got more of a functional flavor. But I've looked at this exact same messaging issue with HTTP a lot and so have some other clever folks, this genuinely seems like the best overall solution to me right now. I would say, give it a shot and try to port, say the CommandEvents package or an example Transactions middleware as a test run. I think you'll find it actually flows pretty naturally when you're implementing.

sagikazarmark:
Ok, I am convinced.

sagikazarmark:
@rosstuck If you choose to merge this PR, let me know, so I can rework the queueing package. Wondering if command events package should also be converted to a middleware package or not.

rosstuck:
@sagikazarmark It's definitely going to get merged, I just need to make time for the last renames and updating the docs (again, le sigh). I'll ping you once the renames are done but before merging so you can work against that. Thanks again! :smile:

rosstuck:
Okay, remaining issues should be resolved here, throwing them out there for comment:

- **CommandBus interface**: I've resplit this into the `CommandBus` interface and `StandardCommandBus` concretion. After seeing this in practice again, I still don't feel it adds anything and I really hate the StandardCommandBus naming. Let me know if anyone's changed their mind on removing it (and no, adding an interface suffix is not an option :smirk_cat: )

- **Closure vs Callable**: I didn't realize this but PHP 5.4 added the ability to execute array or string callables just like a closure, so there's no execution inconsistency there. That makes Callable significantly more flexible than Closure and I recommend we stick with them.

- **Middleware namespacing**: This is changed but likely not as expected. I wanted to move the concretions out of the root but just shuffling everything off to a middleware directory felt off to me. Eventually, I realized the HandlerMiddleware should be under the Handler directory and the Middleware interface should stay top level since it's a core concept to Tactician now. LockingMiddleware is hard but I decided to move it to a generic "Plugins" directory which sorta fits better and we can add any other one-off core, zero deps Middleware there as well.

Lemme know if anyone has any thoughts here, I'm treating this as a Last Call of sorts before moving onto docs which is the sole blocking issue for tagging a release for this package.

sagikazarmark:
I still think the interface should remain in the package. Maybe it's a silly idea, but what if you name the `StandardCommandBus` `Tactician`? I am not entirely sure about your future plans, so if the package itself wants to be part of a bigger architecture, then it might not be a good idea.

I was against `Closure` typehint from the very beginning, `callable`.

If it makes sense, you can call it anything. But what's the plan for plugin packages? Should they go under the plugins namespace?

Also, it is still an open question if the command events implementation should remain a decorator or should become a middleware. It doesn't make much sense to emit a `CommandExecuted` event in the middle of the execution chain. What do you think?

rosstuck:
@sagikazarmark: Hmm, the Tactician name is cool but I do have some larger plans down the road so it might not make sense. I'm also kind of against "themed" names in general since they raise the barrier to entry.

I think the other separate packages should still have their own namespaces like they do now, no change. The plugins namespace here is more for this package's local core plugins.

About the CommandEvents, we could make the HandlerMiddleware call next(), even when it finds and executes a handler. Separate from that, the CommandEvents can do:

```php
public function execute(Command $command, callable $next)
{
    $this->eventEmitter->emit(...); // CommandReceivedEvent
    $returnValue = $next($command);
    $this->eventEmitter->emit(...); // CommandExecutedEvent

    return $returnValue;
}
```

We can add a try/catch block around $next in order to get a commandFailed exception as well.

sagikazarmark:
My problem is that this way a commandExecuted event is emmitted...who knows when in the chain? Executing logic can come even before and after it gets to the event handler part. This is why I think it should remain a decorator, to ensure that the events are truely emitted before any and after all executing logic. IMO it does not make sense to let the dev choose when the events should be emitted by letting it  pushed to the handler stack.

rosstuck:
Provided decorators aren't malicious and do something stupid like running execution logic before calling `$next`, then placing the event dispatching after the `$next` execution is just as reliable as a decorator. In fact, a decorator makes no promises about that either, someone could just as easily write a decorator that performs logic before calling `$this->innerCommandBus->execute($command)`. It's genuinely the exact same structure except innerCommandBus is basically a parameter now, instead of a constructor dependency.

Otherwise, I reckon the fact it can be ordered is a good thing. Maybe I want it before or after a transaction closes or a log files written.

sagikazarmark:
I don't think ordering in this case is a good idea. If you want to rely on events, you should do it using listeners instead of relying on something is added before/after the event middleware. Also we are not talking about decorators in general, but this specific case.

Events should occur at a very specific point in the application. In fact, noone should be able to have control over them.

rosstuck:
@sagikazarmark has suggested moving the Event discussion to https://github.com/thephpleague/tactician-command-events/issues/3 since that discussion is specific.

Anything else before merging? You've got 'till sundown, ya varmints.

sagikazarmark:
:+1: from me

acairns:
Ship it! :)

PR #14, opened 2015-02-02
Feature/improve execptions

rosstuck:


rosstuck:
This was an old feature that got lost between a misdirected PR and the transition to the League repo.

PR #16, opened 2015-02-03
Config improvements

sagikazarmark:
* Added [Editorconfig](http://editorconfig.org/)
* Updated LICENSE year
* Filter specific inclusions instead of exclusions in scrutinizer config
* Adds `travis_retry` to travis config (see [this](http://blog.travis-ci.com/2013-05-20-network-timeouts-build-retries/) for reference)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[composer.json]
+indent_style = space
+indent_size = 4
+
+[*.php]
+indent_style = space
+indent_size = 4
+
+[*.yml]
+indent_style = space
+indent_size = 2
+
+[phpunit.*]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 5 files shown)


rosstuck:
Thanks a ton, nice catch on the License file!

Any objection to setting YAML to 4 spaces so it's consistent? I know Ruby tends to use 2 spaces but a lot of the PHP world uses 4. I'm willing to go through and make it consistent.

sagikazarmark:
Since YAML is not PHP, I think it's better to keep it two spaces. I think that became kind of a standard now. But I don't insist on it, that doesn't really matter.

rosstuck:
Just polled two chat rooms, one was unanimously 4 spaces, the other was all votes for...tabs.

So, I propose 4 spaces here and ultimate vengeance up on the tab people.

sagikazarmark:
Tabs? Those are invalid votes :D ...tabs are invalid in YAML, so they didn't even know what they were saying. :D

sagikazarmark:
Done, but still against it.

rosstuck:
Duly noted and thank you for doing it regardless :)

PR #18, opened 2015-02-04
Make CommandBus final

boekkooi:
With this PR I propose to remove the CommandBus interface and rename the StandardCommandBus to CommandBus and make it final.

*Why remove the interface?*
The main reason is because it adds no value and may give people the idea that other CommandBus types will later be supported by the library

*Why rename StandardCommandBus and make it final?*
Since tactician should focus on one good and simple implementation of a CommandBus aka "there can only be one". 
Also this will avoid strange problems and issues due to people inheriting from CommandBus. 
The way to extend is to use Middleware not inheritance.


rosstuck:
I'm personally inclined to merge this one (the discussion came up before in #10) but after some thinking, I'm tempted to make it final but keep the interface.

One good reason to do this would be allowing full mocking support, something `final` cuts off. [This article](http://ocramius.github.io/blog/when-to-declare-classes-final/) raises some very good points in support of that.

I wouldn't mind seeing StandardCommandBus become just CommandBus again but it's against the rest of the naming convention to use an interface suffix. So, that kind of stinks. We could rename the interface CommandDispatcher or Executor to avoid the issue but that's a bit dodgy.

I'd like to wrap this up and stop the namespace moving by the end of this week though, so if anyone's got thoughts, lemme know.

boekkooi:
@rosstuck I was thinking about the mocking part and the easy solution is to have the Interface I agree with that (then again you could also implement a mock middleware).

So lets then take a look at the CommandBus Interface name. 
So what is this interface describing? I my opinion it is describing how to execute a command so based on that it would be more logical to call the interface CommandExecutor (I would avoid using the name CommandHandler since it will cause confusing with the HandlerMiddleware).

Since tactician currently is only executing a command using a CommandBus it makes total sense to call the interface CommandBus then again will it make sense in the future? and is the goal for tactician not to execute a command?

sagikazarmark:
Well, the only thing I would keep here is the final keyword usage. ;)

I think the CommandBus interface itself describes how a command bus should behave (execute a command). The standard command bus is a hard thing. I proposed to call it Tactician but it might not be a good decision because of future plans.

My other idea is to close on to the problem from an other angle: what is the StandardCommandBus? It is a command bus executing commands using middlewares. Since it IS a command bus (that we now from the implemented interface) the only thing that remains is related to middlewares, so we could play with that name.

rosstuck:
Summing up suggestions:

- CommandBusInterface: Breaks our style guidelines.
- CommandExecutor / CommandProcessor- Introduces another concept that's never really used, folks will still likely typehint CommandBus by default.
- Tactician: "Theme" names are harder to learn and it does mess with some future plans
- MiddlewareCommandBus (or similar): There shouldn't be anything atypical about this, they're first class citizens of Tactician now, so I don't see it worth including in the name.
- MessageBus: Used in SimpleBus as a more generic name but we're never going to support a separate QueryBus and I don't fully believe in this distinction.
- Long term: I've discussed a shared interface with @matthiasnoback, but I don't think any of us want to wait on a PSR. We could create a temporary joint repo with an agreed upon interface.

Fellas, I feel we're shaving a yak here and never gonna get anything perfect. At this point, I'm stumped, tired of rewriting docs and would rather stop moving stuff around and just tag a release people can count on.

If we are gonna do this, I'd vote for the current setup or CommandExecutor. I don't like either but they seem to be the least bad that don't rely on external stuff.

boekkooi:
I'm +1 on CommandExecutor as interface and a final CommandBus.

What is your preference @sagikazarmark?

@rosstuck where is the doc repo I can refactor if for you if you want.

rosstuck:
@boekkooi Docs are using github pages, so they're just on this repo, under the gh-pages branch.

sagikazarmark:
I would leave CommandBus as the interface. That describes a CommandBus
whatever it will be in the future. In this case it doesn't really matter if
middleware is first class citizen or not, the only thing (in naming) that
can make a difference is the implementation itself. So if not something
like that (MiddlewareCommandBus, MiddlewareAcceptor), than I vote for a
generic setup like Standard, FirstClassCitizen, Generic, Basic, Base, etc
CommandBus.

matthiasnoback:
Two notes from my side:
1. We definitely should consider to extract common interfaces and put those in a separate library. Especially when the collection of middlewares would start growing for either of our libraries (I guess Tactician would win that "battle" I think ;)), it would be very useful for users to be able to switch command bus libraries later on.
2. The MessageBus interface originates from the fact that SimpleBus also has an EventBus (no QueryBus, which is something for which I don't think there is a valid use case (yet)) and there was a lot of shared logic. I admit that it's also an experiment - I personally am very happy about it, but a) I wouldn't introduce it to Tactician either, and b) this might make it harder to determine a shared interface like mentioned above.

rosstuck:
@matthiasnoback Yeah, I reckon it's worth doing in the long run. I fear the best route for this might be...*gulp*...PSR. Thoughts?

Regarding this PR, I had a talk with @boekkooi last night and we came to a rough consensus on the following.

- We make StandardCommandBus into plain CommandBus, dropping the interface for now.
- We mark CommandBus as final in the docs but not in the code
- We begin working on a joint set of shared Command Bus interfaces, either external repo or PSR.
- We mark CommandBus final in the code once the shared interfaces are available.

Standalone, these seem like bizarre backwards moves. But what it gives us is:

- A mockable command bus (no enforced final in the short term)
- The ability to extend using middleware, while indicating that extension isn't recommended
- A final naming structure we're happy with, with no real BC breaks.
- We can go ahead and tag a stinkin' release!

We do lose decorators until the shared interfaces are available though. Also, we'll probably have to change `execute` to `handle`. Le sigh.

sagikazarmark:
Dropping the interface doesn't seem to be a good idea. Until having a
common interface, even a CommandBusInterface would be better. That would
allow to simply extend the common interface later without a BC break, so
that it can be used without a major release change.

boekkooi:
@rosstuck I updated the code according to your comment above

rosstuck:
We still have to support a CommandBusInterface in the future though, even after it's become kinda of useless. I took a look around at some other PHP libraries and found this was a similar pattern.

Also, what are the use cases for the interface?
- Decorators: We're encouraging middleware instead
- Mocking: If we don't mark CommandBus as final, it's not a problem
- Alternate implementation: there isn't really an alternate one possible here

So, after a lot of thinking about it, I'm just going to put my foot down and opt for moving StandardCommandBus -> CommandBus, dropping the interface and pushing ahead on a shared interface (PSR or not). I think we can create a shared one in short order with @matthiasnoback, that one can ultimately standalone or extend a future PSR.

That does mean one last related change, we need to rename `execute` to `handle`. We'll also merge #19 as well, changing HandlerMiddleware -> CommandHandlerMiddleware and the namespace will change as well.

At that point, we're not moving or renaming namespaces anymore because we need to tag a release and finalize a structure that our users can count on. Nothing will make us more useless or irrelevant than constantly breaking BC and driving our users nuts. They should be our first concern, and we can always revisit issues before tagging a 1.0 but for now, we need to get this stuff out there to users.

boekkooi:
@rosstuck Code updated with handle rename

PR #19, opened 2015-02-04
Rename HandlerMiddleware to CommandHandler

boekkooi:
I propose to rename HandlerMiddleware to CommandHandler.

The reason to rename HandlerMiddleware is that it in fact is not true middleware since it's a endpoint.
The second reason (and this is a pure personal one) is that the name confused me due to the fact that I expected it to handle middleware.

### DIFF
--- a/examples/1-beginner-standard-usage.php
+++ b/examples/1-beginner-standard-usage.php
@@ -27,7 +27,7 @@ $locator->addHandler(new RegisterUserHandler(), RegisterUserCommand::class);
 
 // Middleware is Tactician's plugin system. Even finding the handler and
 // executing it is a plugin that we're configuring here.
-$handlerMiddleware = new League\Tactician\Handler\HandlerMiddleware(
+$handlerMiddleware = new League\Tactician\Handler\CommandHandlerMiddleware(
     $locator,
     new HandleClassNameInflector()
 );
--- a/examples/2-intermediate-create-middleware.php
+++ b/examples/2-intermediate-create-middleware.php
@@ -59,7 +59,7 @@ class LoggingMiddleware implements Middleware
 //       incoming commands, you're free to do so. Just think carefully!
 
 // Now that we have a working Middleware object, we give it to our CommandBus
-// as a list of middleware. We'll also pass in the HandlerMiddleware we demoed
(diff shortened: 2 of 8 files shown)


rosstuck:
There's a lot of examples out there of Middleware that act as terminal points (even conditionally sometimes). So, in general, I don't see this one happening.

I do understand the naming confusion and we could optionally change the namespace to `League\Tactician\CommandHandler\HandlerMiddleware`, for example. We do get some hilariously long names out of that though, like `League\Tactician\CommandHandler\MethodNameInflector\HandleClassNameInflector`.

boekkooi:
I personally see the namespace option as a good idea but for me this wouldn't solve the problem of having the confusing `HandlerMiddleware` name.

Maybe a better name for the middleware would be `CommandHandlerMiddleware`? or would you rather go funky and rename the namespace and rename `HandlerMiddleware` to `Middleware`?

Let me know what you think but I'm +1 on the namespace rename in any case. Since i believe it's better to type 7 more chars to avoid confusion.

rosstuck:
A couple other folks on chat have voiced support for changing the namespace from `Handler` to `CommandHandler`. I'd support renaming the middleware then to `CommandHandlerMiddleware` since, it's, well, a Middleware that adds CommandHandler support.

Would doing the class rename alone be enough to clarify this or would you still like to have the namespace moved too?

boekkooi:
I feel like the class name renaming is good enough. 
As far as the namespace go's I think `Handler` works better for the locators and inflectors and with renaming the `CommandHandler` to `CommandHandlerMiddleware` i feel that the namespace is clear enough to not be a problem.

P.S. Updated PR

rosstuck:
Lookin' good, thanks for the PR update. :) Going to wait to merge until #18 gets settled and we get all the docs updated in one-go.

rosstuck:
@boekkooi Last question, do you think we'd need to rename HandlerLocator as well to avoid the same confusion?

boekkooi:
@rosstuck I would say no since it is really locating a handler and I personally feel adding command to it won't give any extra value. 
If you aren't happy with the name I would say that renaming the namespace makes it better then renaming the class.

PR #20, opened 2015-02-04
Add conditional middleware example

boekkooi:
Read the code.

### DIFF
--- a/examples/6-conditional-handlers.php
+++ b/examples/6-conditional-handlers.php
@@ -1,0 +1,76 @@
+<?php
+require __DIR__ . '/../vendor/autoload.php';
+require __DIR__ . '/repeated-sample-code.php';
+
+/**
+ * Tactician is meant to be very customizable, it makes no assumptions about
+ * your code.
+ *
+ * For example, let's say you like to handle command X in one way and
+ * command Y in another.
+ *
+ * We can write a custom middleware for that.
+ */
+
+use League\Tactician\Command;
+use League\Tactician\CommandBus;
+use League\Tactician\StandardCommandBus;
+use League\Tactician\Middleware;
+
+// External command
+interface ExternalCommand extends Command
+{
+}
+
+// We create some custom middleware that will forward any ExternalCommand to
+// a separate CommandBus (or HTTP endpoint or queue).
+final class ExternalCommandMiddleware implements Middleware
+{
+    private $commandBus;
+
+    public function __construct(CommandBus $commandBus)
+    {
+        $this->commandBus = $commandBus;
+    }
+
+    public function execute(Command $command, callable $next)
+    {
+        if ($command instanceof ExternalCommand) {
+            return $this->commandBus->execute($command);
+        }
+        return $next($command);
+    }
+}
+
+// and we'll create a custom command handler/middleware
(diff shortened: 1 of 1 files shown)


rosstuck:
Legit pattern! Can you give a more specific use case for why you'd do this though? I feel in practice you wouldn't go bus->bus very often, but more bus->handler or bus->queue or bus->network.

boekkooi:
@rosstuck One of the cases where I used this in the past was when there was one "main" CommandBus that was used for multiple type of commands. 

For example the default command is using db connection 1 the second command type is using db connection 2 and the final is using a external message queue. So for the DB connections we needed to add transaction and locking but for the message queue we just needed to log and transform the command and add it to a message queue.

This can be done using a handler but that would mean that you can't take advantage of default middleware so the bus->bus implementation is the best one for these cases. It's a minimal overhead for the maximum gain.

Hope this makes sence.

rosstuck:
Hmm, would you be willing to rename the example to Command Routing instead of Conditional Middleware? I totally see that you can do both with this and think folks would get it, it might just zero in on the end goal more.

examples/6-conditional-handlers.php
```diff
@@ -0,0 +1,82 @@
...
+use League\Tactician\StandardCommandBus;
+use League\Tactician\Middleware;
+
+// External command
+interface ExternalCommand extends Command
+{
+}
+
+// We create some custom middleware that will forward any ExternalCommand to
+// a separate CommandBus.
```
rosstuck: Suggested text addition: "...or HTTP endpoint or queue." Make certain they realize it doesn't have to be another CommandBus

examples/6-conditional-handlers.php
```diff
@@ -0,0 +1,82 @@
...
+final class ExternalCommandMiddleware implements Middleware
+{
+    /**
+     * @var CommandBus
+     */
+    private $commandBus;
+
+    /**
+     * @param CommandBus $commandBus
+     */
```
rosstuck: You can probably drop the docblocks here for brevity. It's pretty clear from the naming and typehints.

examples/6-conditional-handlers.php
```diff
@@ -0,0 +1,82 @@
...
+class MyExternalCommand implements ExternalCommand
+{
+}
+
+// Now  let's instantiate our ExternalCommandHandler, StandardCommandBus and ExternalCommandMiddleware
+$externalCommandHandler = new ExternalCommandHandler();
+$externalCommandBus = new StandardCommandBus([$externalCommandHandler]);
+$externalMiddleware = new ExternalCommandMiddleware($externalCommandBus);
+
+// Time to create out main CommandBus
```
rosstuck: Our, not out. :smile:

examples/6-conditional-handlers.php
```diff
@@ -0,0 +1,82 @@
...
+$commandBus = new \League\Tactician\StandardCommandBus([$externalMiddleware, $handlerMiddleware]);
```
rosstuck: FQCN here but it's already imported via use at the top.

boekkooi:
@rosstuck PR is updated

PR #28, opened 2015-02-12
Adds event middleware to the docs

sagikazarmark:


plugins/event-middleware.md
```diff
@@ -0,0 +1,62 @@
...
+use League\Tactician\CommandEvents\EventMiddleware;
+
+$emitter = new Emitter;
+$eventMiddleware = new EventMiddleware($emitter);
+
+// other possible solution
+// $eventMiddleware->setEmitter($emitter);
+~~~
+
+You can also catch an error and prevent it from causing the application fail:
```
rosstuck: Little typo: "prevent it from causing the application **to** fail:"

plugins/event-middleware.md
```diff
@@ -0,0 +1,62 @@
...
+title: Event Middleware
+---
+
+# Event Middleware
+
+[![Author](http://img.shields.io/badge/author-@sagikazarmark-blue.svg?style=flat-square)](https://twitter.com/sagikazarmark)
+[![Source](http://img.shields.io/badge/source-league/tactician--command--events-blue.svg?style=flat-square)](https://github.com/thephpleague/tactician-command-events)
+[![Packagist](http://img.shields.io/packagist/v/league/tactician-command-events.svg?style=flat-square)](https://packagist.org/packages/league/tactician-command-events)
+
+This plugin lets you to listen to some events emitted during command execution:
```
rosstuck: Little typo: "lets you ~~to~~ listen"

rosstuck:
Looking good! One last request: can you change the code samples to be inline with PSR-2? Specifically adding () brackets after "new ClassName" and wrapping the functions onto new lines? I don't always like it either but I reckon it's better to be consistent in the docs.

sagikazarmark:
Is it in the PSR? Having () after class instantiation even if there is no argument list? I can't find any clue for that. If it is explicitly in the standard then so be it. Honestly, It rather bothers me.

sagikazarmark:
> wrapping the functions onto new lines?

I am not sure what you mean.

rosstuck:
Paranthese are definitely the expected style and used everywhere else, so please add them. :)

PSR-2 does definitely require a closure being passed like this

```php
$eventMiddleware->addListener(
    'command.failed',
    function (CommandFailed $event) {
        $event->catchException(); // without calling this the exception will be thrown
    }
);
```

(Also, note the space after `function` and before the opening bracket)

sagikazarmark:
Done. Yeah, the space is a good point, didn't notice.

PR #32, opened 2015-02-16
Adds Bernard plugin docs

sagikazarmark:
It uses the new `tactician-bernard` in links, so they won't work until the package rename is done.

### DIFF
--- a/plugins/bernard.md
+++ b/plugins/bernard.md
@@ -1,0 +1,54 @@
+---
+layout: default
+permalink: /plugins/bernard/
+title: Bernard
+---
+
+# Bernard
+
+[![Author](http://img.shields.io/badge/author-@sagikazarmark-blue.svg?style=flat-square)](https://twitter.com/sagikazarmark)
+[![Source](http://img.shields.io/badge/source-league/tactician--bernard-blue.svg?style=flat-square)](https://github.com/thephpleague/tactician-bernard)
+[![Packagist](http://img.shields.io/packagist/v/league/tactician-bernard.svg?style=flat-square)](https://packagist.org/packages/league/tactician-bernard)
+
+
+This plugin provides you tools for both sending commands to and consuming from a queue, so you should install it in both your application and (if you have any) dedicated worker package/instance.
+
+You can run your consumer from your application environment, however you should avoid it if possible. Bootstrapping your application consumes more resource than necessary. (That said, sometimes it is simply easier.) In this case you have to make sure that you provide your consumer instance with all the data it needs: database connection details, required dependencies, etc.
+
+
+### Remote execution
+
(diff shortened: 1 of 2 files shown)


plugins/bernard.md
```diff
@@ -0,0 +1,54 @@
...
+# Bernard
+
+[![Author](http://img.shields.io/badge/author-@sagikazarmark-blue.svg?style=flat-square)](https://twitter.com/sagikazarmark)
+[![Source](http://img.shields.io/badge/source-league/tactician--bernard-blue.svg?style=flat-square)](https://github.com/thephpleague/tactician-bernard)
+[![Packagist](http://img.shields.io/packagist/v/league/tactician-bernard.svg?style=flat-square)](https://packagist.org/packages/league/tactician-bernard)
+
+
+This plugin provides you tools for both sending commands to and consuming from a queue, so you should install it in both your application and (if you have any) dedicated worker package/instance.
+
+You can run your consumer from your application environment, however you should avoid it if possible. Bootstrapping your application consumes more resource than necessary. (That said, sometimes it is simply easier.) In this case you have to make sure that you provide your consumer instance with all the data it needs: database connection details, required dependencies, etc.
```
rosstuck: As mentioned elsewhere, I'm not in complete agreement with recommending this to people. These days, "application environment" usually means their DI container and CLI library, which I think they _should_ bootstrap with.

sagikazarmark: @rosstuck the whole point of queue-worker architectures is that your workers can be ran EVERYWHERE. That's why they are scalable.

Imagine the scenario: you are running a blazing fast webserver with an ecommerce application and 10000 visitors a day. Everyone buys something, wants to pay with credit card, which means you have to generate an invoive for them. If you are smart enough, you generate a PDF as well, so they can download it. Now imagine that you have to generate every PDF file, with signing and storing/uploading/sending out in email whatever on the same webserver. It will obviously slow things down. That's why workers should be moved to a separate server.

SO the point is to have workers completely separated from the application as they are NOT part of the *application environment*.

rosstuck:
As stated elsewhere, the physical location of the code (this server or another) is irrelevant to the configuration since you're almost certainly using the same codebase. Further, the hallmark of queues is horizontal scalability. Trying to eek out more performance from the same server is more akin to vertical scalability.

Look, I'm familiar with queueing and I agree that people should think carefully about the memory resources they boot. But it's not like folks are starting up their web stack in their workers.

Finally, the Tactician docs should focus on setting up our plugins and tooling, not giving high-level advice on building distributed systems: it's just out-of-scope for our docs.

sagikazarmark:
Your points on the docs are valid, I am going to correct that. Let's move the rest of the discussion to the relevant issue.

PR #38, opened 2015-03-13
Improves travis config

sagikazarmark:


.travis.yml
```diff
@@ -3,10 +3,17 @@ language: php
...
     - hhvm
 
+matrix:
+    include:
+        - php: 5.5
+          env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
+
 before_script:
-    - travis_retry composer install --no-interaction
+    - travis_retry composer self-update
```
GrahamCampbell: :-1:

GrahamCampbell:
Also, this "improvement" doesn't actually work...

.travis.yml
sagikazarmark: What's the problem with it?

The reason I usually leave it in is that travis rarely update the composer executable. I had several issues when the `^` was introduced.

rdohms: As far as I know its a good practice to update composer before running a build, sure it may add a few extra seconds, but it gets you a decent version of composer.

rdohms:
I think running a build with `--prefer-lowest` is a good idea.
Since the lock is not committed switching `install` for `update` literally makes no difference so this implementation is good.

I would however bump the version for phpcs to `~2.3` to avoid the CS error that shows up in the build.
I would also add PHP7 and HHVM to `allow_failures` (http://docs.travis-ci.com/user/build-configuration/#Rows-That-are-Allowed-To-Fail) unless its the libraries intent to be compatible with this.

Also i would recommend adding back the self-update command as you can notice composer is throwing warnings.

sagikazarmark:
`update` is needed for `--prefer-lowest`. Otherwise it doesn't work.

I think it is getting a League standard to support hhvm and 7.0 and only allow failure if it is really needed. Currently the build passes for borh.

rdohms:
@sagikazarmark yes, i'm just telling @rosstuck that he should not worry about switching install to update as it renders the same results.

As for League standards, i do not know what they are, but the fact that "it passes now" is not really indicative of future compatibility, Either way.

sagikazarmark:
@rdohms oh, I see.

Also, I fixed the CS issue (added the extra space). I think `~2.0` installs `2.3` if it is released, so I see no point in explicitly bumping to that version.

.travis.yml
GrahamCampbell: > As far as I know its a good practice to update composer before running a build

No, quite the reverse. It's a BAD PRACTICE. Travis DO NOT RECOMMEND it. That's why they DO NOT do it for you...

rdohms: Got a link?

As far as I see it its bad, this version may not even have caret support yet, depending on how old it is. @seldaek recommends always updating composer to the latest since its not stable yet, i'll stick with him.

sagikazarmark: Travis only says it updates it every 30-60 days, so you might see a warning. But nothing about such a recommendation...

GrahamCampbell: > Got a link?

> But nothing about such a recommendation...

Yep, I've got a link. One sec...

GrahamCampbell: https://github.com/travis-ci/travis-build/pull/384

GrahamCampbell: That means the official decision from travis was not to update composer.

sagikazarmark: I can't see any advisory NOT to update it manually (actually the against). Also, the resolution was the warning I've already mentioned. So NOT doing something is not the same as not recommended.

rdohms: All I see is you bullying a guy with harsh words and him retracting his own suggestion, I see no official position from Travis there. Unless you work for Travis and i'm missing that information.

GrahamCampbell: > All I see is you bullying a guy with harsh words and him retracting his own suggestion

No, we had a discussion. I said I thought it was a bad idea. It was his decision to close the pull. I didn't make him do anything.

sagikazarmark: I agree with @rdohms, you was a bit more "aggressive" than you should have be. Apart from that, the idea was to update, then to suggest a manual update. Actually the warning is rather a "do-what-you-want" solution.

boekkooi:
Some information:
- Travis says that the composer time warning can be [`safely ignored`](http://docs.travis-ci.com/user/languages/php/#Installing-Composer-packages). 
- The composer `--prefer-lowest` was added to avoid BC breaks and bugs. Take a look at [Testing your composer dependencies with prefer-lowest](http://evertpot.com/testing-composer-prefer-lowest/).

So personally I think this PR in it's current state is fine except for not allowing PHP 7 to fail.
Adding a [`allow_failures`](http://docs.travis-ci.com/user/build-configuration/#Rows-That-are-Allowed-To-Fail) for PHP 7 should fix this.
```YAML
matrix:
  allow_failures:
    - php: 7
```

.travis.yml
GrahamCampbell: Ultimately, this isn't worth getting worked up over. This is sooo trivial.

If you want to self-update composer on your project, then do it. The world isn't going to implode. I'm only recommending that you don't do it.

sagikazarmark:
It is not the composer warning I am concerned about, but the added features that take two months to make into composer. People (hopefully) always use the latest version, travis should as well. However it doesn't so it makes sense to update composer (especially if you started to use the carret and --prefer-lowest). I think it saves us from trouble when you start using the next new feature, it will just work on travis as well.

About the allow failures thing. Currently (in theory) there is no BC break in 7.0 so it should just with it as well. So I think if it works in the current state, it will probably work in the future as well.

It is time for @rosstuck to add his points.

rosstuck:
I'm laid up sick so I asked @rdohms (aka The Dohmsraken) to take a look at this since it seems like a good idea. I'm in favor of merging it (with the composer update).

I agree with @boekkooi and @rdohms about adding PHP 7 to the allowed_failures (for now) but HHVM is a target I'm deliberately supporting, so let's leave that as breaking. I'm actually curious about getting a League recommendation or guideline for the support matrix (more a rule of thumb than a hard rule) but that's an issue for the mailing list or IRC perhaps.

Let's also lift the PHPCS requirement to 2.3 as well, to ensure that dev environments are consistent even when deliberately forcing lower versions in QA.

Thank you to everyone for your contributions, they are greatly appreciated.

@GrahamCampbell I appreciate your input; however, please moderate your tone in this issues (and others) better. I would like Tactician to be a place where newcomers can feel welcome to contribute and receive coaching. What they see in existing issues is important for that reason. I hope you would like to continue contributing but I will require this going forward. Thank you.

sagikazarmark:
I don't insist on having PHP 7 as breaking, this is what I referred to:

> "If its failing feel free to put it in the "allowed failures" list, but try to get it working as soon as you can. There aren't any BC breaks that should make our life hell as the code is all new enough to not having mysql_query or any of that bullshit." - by @philsturgeon

sagikazarmark:
Damn, you are right, I just pasted your example. ;)

sagikazarmark:
@rosstuck IMO I did everything you asked, if you are fine with it I squash the commits into one.

boekkooi:
:thumbsup:

rosstuck:
Go for the squash if you like, I'm fine with merging as is. :)

sagikazarmark:
Done

rosstuck:
@sagikazarmark Thank you for contributing. :)

PR #46, opened 2015-03-31
Adds command name extractors and named command

sagikazarmark:
This PR adds CommandNameExtractors which help to identify the name which can be used in locators.

I have three ways in mind on implementing them, two of them are not BC breaking, but not really nice as well.

1. Create a `CommandNameExtractorAware` trait for locators, and return a default (ClassNameExtractor) if none is set. 

2. This is an extension of the above. An extractor is optionally passed to the CommandHandlerMiddleware (the default is created if none passed). The extractor is set like described above.

3. The locator interface is "changed" to accept the command name (IMO no actual change is required). An extractor should be passed to the middleware, and only the command name should be passed to the locator.

I would say BC breaking is not that much in this case (dev must pass another argument, DIC solves it in most cases anyway)

@rosstuck I didn't want to implement all of them, so whichever way we choose, I will implement it in this PR.

### DIFF
--- a/src/Handler/CommandNameExtractor/CommandNameExtractor.php
+++ b/src/Handler/CommandNameExtractor/CommandNameExtractor.php
@@ -15,5 +15,5 @@ interface CommandNameExtractor
      *
      * @return string
      */
-    public function getNameForCommand($command);
+    public function extract($command);
 }
--- a/src/Handler/CommandNameExtractor/CommandNameExtractor.php
+++ b/src/Handler/CommandNameExtractor/CommandNameExtractor.php
@@ -2,6 +2,8 @@
 
 namespace League\Tactician\Handler\CommandNameExtractor;
 
+use League\Tactician\Exception\CanNotDetermineCommandNameException;
+
 /**
  * Extract the name from a command so that the name can be determined
  * by the context better than simply the class name
@@ -14,6 +16,8 @@ interface CommandNameExtractor
      * @param object $command
      *
      * @return string
+     *
+     * @throws CannotDetermineCommandNameException
      */
     public function extract($command);
 }
--- a/src/Handler/CommandNameExtractor/CommandNameExtractor.php
+++ b/src/Handler/CommandNameExtractor/CommandNameExtractor.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace League\Tactician\Handler\CommandNameExtractor;
+
+/**
+ * Extract the name from a command so that the name can be determined
+ * by the context better than simply the class name
+ */
+interface CommandNameExtractor
+{
+    /**
+     * Extract the name from a command
+     *
+     * @param object $command
+     *
+     * @return string
+     */
+    public function getNameForCommand($command);
+}
--- a/src/Handler/CommandNameExtractor/NamedCommand.php
+++ b/src/Handler/CommandNameExtractor/NamedCommand.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace League\Tactician\Handler\CommandNameExtractor;
+
+/**
+ * Exposes a name for a command
+ */
+interface NamedCommand
+{
+    /**
+     * Returns the name of the command
+     *
+     * @return string
+     */
+    public function getName();
+}
--- a/src/Handler/CommandNameExtractor/NamedCommand.php
+++ b/src/Handler/CommandNameExtractor/NamedCommand.php
@@ -12,5 +12,5 @@ interface NamedCommand
      *
      * @return string
      */
-    public function getName();
+    public function getCommandName();
 }
--- a/src/Handler/CommandNameExtractor/NamedCommandExtractor.php
+++ b/src/Handler/CommandNameExtractor/NamedCommandExtractor.php
@@ -13,7 +13,7 @@ class NamedCommandExtractor implements CommandNameExtractor
     public function extract($command)
     {
         if ($command instanceof NamedCommand) {
-            return $command->getName();
+            return $command->getCommandName();
         }
 
         // fallback to class name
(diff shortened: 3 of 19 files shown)


rosstuck:
Let's do option 3 and just get it over with.

src/Handler/CommandNameExtractor/CommandNameExtractor.php
```diff
@@ -0,0 +1,19 @@
...
+interface CommandNameExtractor
+{
+    /**
+     * Extract the name from a command
+     *
+     * @param object $command
+     *
+     * @return string
+     */
+    public function getNameForCommand($command);
```
rosstuck: Pendantic but I'd prefer `getNameOfCommand` or `extract` to mirror the interface better.

src/Handler/CommandNameExtractor/NamedCommand.php
```diff
@@ -0,0 +1,16 @@
...
+ * Exposes a name for a command
+ */
+interface NamedCommand
+{
+    /**
+     * Returns the name of the command
+     *
+     * @return string
+     */
+    public function getName();
```
rosstuck: `getName()` is a really common method on a command so we should change it to something less likely to collide with user params. `getCommandName()` seems pretty straightforward.

src/Handler/CommandNameExtractor/NamedCommandExtractor.php
```diff
@@ -0,0 +1,22 @@
...
+     * {@inheritdoc}
+     */
+    public function getNameForCommand($command)
+    {
+        if ($command instanceof NamedCommand) {
+            return $command->getName();
+        }
+
+        // fallback to class name
+        return get_class($command);
```
rosstuck: I'd rather it just throw an exception here, honestly. If you'd like to have a custom fallback scenario, I'd recommend doing that per app or creating a chained version.

We might want a standard `CouldNotDetermineCommandName` exception that all extractors can throw for situations like this.

src/Handler/CommandNameExtractor/NamedCommandExtractor.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace League\Tactician\Handler\CommandNameExtractor;
```
rosstuck: After some thought, rather than placing the NamedCommandExtractor and NamedCommand interface under the Handler namespace where they get a bit separated, it would be clearer to place them under a `Plugins\NamedCommand` namespace.

They'd still ship with the core packages but putting the two parts together under the same folder really improves the cohesion and the plugin aspect hints that they're an optional setup, rather than people assuming they need to implement a NamedCommand interface always.

src/Handler/CommandNameExtractor/NamedCommandExtractor.php
sagikazarmark: Exception name variations (based on current exceptions):

- CanNotDetermineCommandName
- CanNotDetermineCommandNameException
- CommandNameCouldNotBeDetermined

(Just to make sure we remain consistent. Also, I am pretty against the `Exception` suffix, especially if we place exceptions under a separate namespace.

sagikazarmark: That said, it is your call.

src/Handler/CommandNameExtractor/NamedCommandExtractor.php
sagikazarmark: I think it makes sense. but `...\NamedCommand\NamedCommand`? Isn't that weird?

src/Handler/CommandNameExtractor/NamedCommandExtractor.php
sagikazarmark: Also, I am thinking about adding a fallback here anyway. I mean, probably not all commands implement the NamedCommand interface. So this extractor could accept another in the constructor (optionally) and if it is there, the extractor could use that instead of throwing an exception

rosstuck: The most consistent would be CanNotDetermineCommandNameException, let's just run with that.

I'd still be against against the fallback here. Most of the time, folks will choose one strategy and stick with it throughout their app. I think the setup makes it easy to do custom strategies and that's totally alright (and supported!). I'd just rather see folks create their own strategies than making this part more complex.

src/Handler/CommandNameExtractor/NamedCommandExtractor.php
rosstuck: On a personal level, I don't think so. Having everything in a common namespace / folder really brings them together.

We've got to name the folder though, so we often assign it a name given to the most important concept, which being important, is also implemented as an object or interface inside that folder. Cheesy example: If I was doing an email library, I might have `Foo\Email\Sender` and `Foo\Email\Recipient` as value objects, but I'd probably assign them both to `Foo\Email\Email`...cause it's an actual concept too.

sagikazarmark: Ok, convinced me

src/Handler/CommandNameExtractor/NamedCommandExtractor.php
sagikazarmark: Ok, I won't need this feature. I just thought about it, because of the single-command-bus "priniciple". In that case you have to make ALL your commands a named command.

sagikazarmark:
@rosstuck Any chance you can look into this during the weekend? I have to put a project into staging the next week and I would be happy if I could avoid messing with composer repositories.

rosstuck:
Working on it now, while planning easter party. Goal is to get 0.5 out this weekend. :)

sagikazarmark:
Oh, didn't think we get that far, but :100: :+1:

sagikazarmark:
Awesome :+1:

rosstuck:
Gonna do some more local reading and might make a few tweaks but overall, this looks great. Fantastic work and thank you. :+1:

rosstuck:
Also, I'll handle the docs so no worries about that. :)

sagikazarmark:
Thanks

PR #48, opened 2015-04-02
Using composer to register dev namespaces and load classes on global namespace

renan:
No need for autoloading hacks, as described in the former `bootstrap.php` file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,14 @@
             "League\\Tactician\\": "src"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "League\\Tactician\\Tests\\": "tests"
+        },
+        "files": [
+            "tests/Fixtures/Command/CommandWithoutNamespace.php"
+        ]
+    },
     "extra": {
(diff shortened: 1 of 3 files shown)


rosstuck:
That's clever, I like it! However, I showed it to the Dohmsraken who mentioned that it would probably autoload tests + fixtures for every project that just depended on Tactician. I'm not quite sure that's what we want so, so I'm going to investigate this weekend and get back to you.

Sorry for the wait but I really appreciate the PRs!

renan:
Only on development, probably yes. Perhaps rename `CommandWithoutNamespace` to `TacticianCommandWithoutNamespace` so it doesn't risk conflicting with other classes?

sagikazarmark:
@renan is right, those files would only be loaded when you develop tactician (use composer in the tactician package), which is exactly what we want.

rosstuck:
Yup, confirmed on the composer docs. Looks good!

rosstuck:
@renan Thank you! :)

PR #49, opened 2015-04-02
Dockblock and other code cleaning changes

sagikazarmark:
Most important changes:

- Uses {@inheritdoc} where possible
- Improves readability
- Eliminates `/**final**/`, uses @final dockblock instead
- "Common" core PHP classes (eg. Closure) are not imported to the current namespace

### DIFF
--- a/src/Handler/Locator/InMemoryLocator.php
+++ b/src/Handler/Locator/InMemoryLocator.php
@@ -62,10 +62,8 @@ class InMemoryLocator implements HandlerLocator
     }
 
     /**
-     * Retrieve handler for the given command
+     * {@inheritdoc}
      *
-     * @param object $command
-     * @return object
      * @throws MissingHandlerException
      */
     public function getHandlerForCommand($command)
--- a/src/Plugins/LockingMiddleware.php
+++ b/src/Plugins/LockingMiddleware.php
@@ -23,9 +23,7 @@ class LockingMiddleware implements Middleware
     /**
      * Queues incoming commands until the first has completed
      *
-     * @param object $command
-     * @param callable $next
-     * @return mixed
+     * {@inheritdoc}
      */
     public function execute($command, callable $next)
     {
--- a/src/CommandBus.php
+++ b/src/CommandBus.php
@@ -2,16 +2,17 @@
 
 namespace League\Tactician;
 
-use Closure;
 use League\Tactician\Exception\InvalidCommandException;
 
 /**
  * Receives a command and sends it through a chain of middleware for processing.
+ *
+ * @final
  */
-/* final */class CommandBus
+class CommandBus
 {
     /**
-     * @var Closure
+     * @var \Closure
      */
     private $middlewareChain;
 
@@ -27,6 +28,7 @@ use League\Tactician\Exception\InvalidCommandException;
      * Executes the given command and optionally returns a value
      *
      * @param object $command
+     *
      * @return mixed
      */
     public function handle($command)
@@ -41,7 +43,8 @@ use League\Tactician\Exception\InvalidCommandException;
 
     /**
      * @param Middleware[] $middlewareList
-     * @return Closure
+     *
+     * @return \Closure
      */
     private function createExecutionChain($middlewareList)
     {
--- a/src/Exception/CanNotInvokeHandlerException.php
+++ b/src/Exception/CanNotInvokeHandlerException.php
@@ -18,6 +18,7 @@ class CanNotInvokeHandlerException extends \BadMethodCallException implements Ex
     /**
      * @param object $command
      * @param string $reason
+     *
      * @return static
      */
     public static function forCommand($command, $reason)
--- a/src/Exception/InvalidCommandException.php
+++ b/src/Exception/InvalidCommandException.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace League\Tactician\Exception;
 
 /**
@@ -13,6 +14,7 @@ class InvalidCommandException extends \RuntimeException implements Exception
 
     /**
      * @param mixed $invalidCommand
+     *
      * @return static
      */
     public static function forUnknownValue($invalidCommand)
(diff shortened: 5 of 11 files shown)


src/Handler/Locator/InMemoryLocator.php
```diff
@@ -62,10 +62,8 @@ protected function addHandlers(array $commandClassToHandlerMap)
     }
 
     /**
-     * Retrieve handler for the given command
+     * {@inheritdoc}
```
GrahamCampbell: :-1: This isn't very useful. Docblocks should be human readable.

src/Plugins/LockingMiddleware.php
```diff
@@ -23,9 +23,7 @@ class LockingMiddleware implements Middleware
     /**
      * Queues incoming commands until the first has completed
      *
-     * @param object $command
-     * @param callable $next
-     * @return mixed
+     * {@inheritdoc}
```
GrahamCampbell: :-1: This isn't very useful. Docblocks should be human readable.

GrahamCampbell:
Aside from the inheritdoc changes, :+1:

sagikazarmark:
If I am correct, the inheritdoc tag is supposed to indicate inherited text documentation, not details of the function definition, thus this is the proper usage of it. However, this tag is unofficially recommended for indicating function documentation inheritance.

In both cases the inheritdoc tag is used to indicate some sort of inheritance. Also, this is just a consistency change, inheritdoc is used everywhere else.

So I can't decide whether you say this is improper usage (in which case check the first paragraph) or you are ultimately against inheritdoc (in which case see second paragraph for answer).

GrahamCampbell:
> So I can't decide whether you say this is improper usage

I'm just saying that I never use inheritdoc because it's not human readable, and for me, that's very important for **documentation**.

sagikazarmark:
Well, I see the point. However if the code is correctly documented (eg. has an own website, which is the case now) then in-code documentation is rather for developers (working on the code) than "users".

As said, inheritdoc is generally used in the code.

If I am correct @rosstuck has some plans to improve in-code documentation (more like ruby doc), so this will probably change in the future.

This PR is mostly about removing inconsistency.

mvriel:
Hi people!

@rosstuck has asked me to provide input on the `@inheritDoc` tag. Please note that the following is my personal opinion and not a recommendation by the phpDocumentor project.

> Full disclosure: for those who don't know; I am the project lead for phpDocumentor and people tend to think I know stuff about DocBlocks ;)

## TL;DR

I prefer `{@inheritDoc}` over omitting the DocBlock but I prefer writing a complete DocBlock over `{@inheritDoc}`. 

## The details

I personally prefer to write my DocBlocks in full and only use `@inheritDoc` when I am too lazy or do not see the added value in copy-pasting a DocBlock from a parent class. 

Writing a DocBlock in full has the benefit that you do not have coupling between your class hierarchy and allows your developer to instantly read the documentation instead of having to open the parent class file (which may be located somewhere else and can be a nuisance).

Using IDEs and phpDocumentor will significant reduce this burden as that recognize the `{@inheritDoc}` tag and resolve the docs, but for casual reading on Github and when using Sublime Text or VIM this can be suboptimal.

Using `{@inheritDoc}` is nonetheless preferred to omitting a DocBlock. Technically speaking a DocBlock is automatically (partially) inherited when it is omitted but `{@inheritDoc}` makes it explicit that the element is intentionally not documented in more detail but the reader is deferred to the parent class (or interface).

This also uncovers another disadvantage to `{@inheritDoc}` and inheritance: not all tags are automatically inherited when using IDEs and tools. I refer to http://www.phpdoc.org/docs/latest/guides/inheritance.html for more details on this matter.

_Another interesting detail to note is that the usage of `{@inheritDoc}` is actually incorrect as this is an inline tag that may officially only be used in a Description of a DocBlock and cannot be used as the only contents of a DocBlock. This inline tag does nothing more than include the contents of the parent DocBlock's Description and not the summary or tags. Because Doctrine and Symfony have started using this style it is now adopted as an unofficial extension of the DSL but in the future a new tag `@inheritDoc` (without braces) will be introduced in the PHPDoc language to fulfil the same role using a valid syntax._

rosstuck:
Thank you @mvriel. Looking at the specific situation here, with the `InMemoryLocator` and `LockingMiddleware`, I would probably do better to add very specific, concrete descriptions there.

Therefore, I'd propose merging this as it stands with the idea the inheritdocs work as a TODO for me to write better docs tomorrow when I have some free time. :)

rosstuck:
@sagikazarmark I just merged a separate docblock improvement PR before this one (unrelated), so I can no longer merge. Will resolve the conflict tomorrow. :)

sagikazarmark:
> Another interesting detail to note is that the usage of {@inheritDoc} is actually incorrect as this is an inline tag that may officially only be used in a Description of a DocBlock and cannot be used as the only contents of a DocBlock.

This is what I referred earlier.

@rosstuck Don't worry (be happy, ooooh....) about the conflict, I will resolve it.

My particular concern about writting EXACTLY THE SAME documentation in every implementation of an interface is that it can easily get inconsistent. Unit tests will fail if you change the interface, but not the implementations. But they won't fail if you forgot to update the docs. Also, in case of usual interface-implementations, you are documenting the interface, not the class itself. If there is a difference between the interface and the implementation (exception throw, optional parameter) then of course it should be included in the implementation's docs.

sagikazarmark:
> but in the future a new tag @inheritDoc (without braces) will be introduced in the PHPDoc language to fulfil the same role using a valid syntax.
@mvriel This is what I am really-really-really looking forward to.

src/Plugins/LockingMiddleware.php
GrahamCampbell: So are these getting removed then?

sagikazarmark: Not now, but as @rosstuck said, he will improve docs when he has time. Also, @mvriel pointed that there will be an official inheritDoc tag, which I think will be used in the league (as the curly-braced version is pretty common nowadays).

PR #50, opened 2015-04-04
Updates branch alias

sagikazarmark:


sagikazarmark:
@rosstuck This one is also needed to use tactician with version constraints and not dev.master.

sagikazarmark:
Thanks

Issue #54, opened 2015-04-05
Assert that array contains only Middleware


boekkooi:
We should check this on https://github.com/thephpleague/tactician/blob/master/src/CommandBus.php#L24 or https://github.com/thephpleague/tactician/blob/master/src/CommandBus.php#L55

I know it's defensive but better to catch a stupid mistake when creating the chain then when executing it.

rosstuck:
Good point, will definitely pick this up for the next release!

PR #56, opened 2015-04-06
Make sure that only valid middlewares are accepted

rtuin:
This should cover the issue found in https://github.com/thephpleague/tactician/issues/54

### DIFF
--- a/src/Exception/InvalidMiddlewareException.php
+++ b/src/Exception/InvalidMiddlewareException.php
@@ -2,7 +2,10 @@
 
 namespace League\Tactician\Exception;
 
-class InvalidMiddlewareException extends \LogicException implements Exception
+/**
+ * Thrown when the CommandBus was instantiated with an invalid middleware object
+ */
+class InvalidMiddlewareException extends \InvalidArgumentException implements Exception
 {
     public static function forMiddleware($middleware)
     {
--- a/src/Exception/InvalidMiddlewareException.php
+++ b/src/Exception/InvalidMiddlewareException.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace League\Tactician\Exception;
+
+class InvalidMiddlewareException extends \LogicException implements Exception
+{
+    public static function forMiddleware($middleware)
+    {
+        $name = is_object($middleware) ? get_class($middleware) : gettype($middleware);
+        $message = sprintf(
+            'Cannot add "%s" to middleware chain as it does not implement the Middleware interface.',
+            $name
+        );
+        return new static($message);
+    }
+}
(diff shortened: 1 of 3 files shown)


tests/CommandBusTest.php
```diff
@@ -67,4 +68,24 @@ public function testHandleThrowExceptionForInvalidCommand()
...
     }
+
+    public function testIfOneCanOnlyCreateWithValidMiddlewares()
+    {
+        $middlewareList = [$this->getMock('stdClass')];
+
+        try {
+            new CommandBus($middlewareList);
+            $this->fail('Should not be able to create a CommandBus without a valid list of middlewares!');
+        } catch (InvalidMiddlewareException $caught) {
```
boekkooi: There is no need to use a try catch seehttps://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions also using a dataprovider to test for a string or null maybe better

rtuin: This is how I like to test for expected exceptions, as this allows me to present a meaningful message when the exception does not occur.

boekkooi: It is still possible since the exception check is done once the is finished.
```PHP
$this->setExpectedException('League\Tactician\Exception\InvalidMiddlewareException');

$middlewareList = [$this->getMock('stdClass')];
new CommandBus($middlewareList);
$this->fail('Should not be able to create a CommandBus without a valid list of middlewares!');
```

src/Exception/InvalidMiddlewareException.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace League\Tactician\Exception;
+
+class InvalidMiddlewareException extends \LogicException implements Exception
```
rosstuck: Needs a little docblock and maybe `InvalidArgumentException` is a better base class here?

src/Exception/InvalidMiddlewareException.php
```diff
@@ -0,0 +1,16 @@
...
+{
+    public static function forMiddleware($middleware)
+    {
+        $name = is_object($middleware) ? get_class($middleware) : gettype($middleware);
```
rosstuck: Good move. :)

tests/CommandBusTest.php
rosstuck: Huh, TIL.

The exception itself is already fairly meaningful so just the expectedException annotation is enough for me here. If you'd like to be more specific, either format is fine with me.

rosstuck:
Gimmie a sign when the comments are addressed and I'll gladly merge this one. Thanks a ton! :)

tests/CommandBusTest.php
rtuin: Hah, interesting indeed! I cleaned up the test method a bit accordingly :+1:

rtuin:
@rosstuck 
image

rosstuck:
Thank you!

Issue #58, opened 2015-04-11
CommandBus Builder


Kilte:
What do you think about adding a simple builder?

For example:
```php
<?php

use League\Tactician\CommandBus;
use League\Tactician\Handler\CommandHandlerMiddleware;
use League\Tactician\Handler\CommandNameExtractor\ClassNameExtractor;
use League\Tactician\Handler\CommandNameExtractor\CommandNameExtractor;
use League\Tactician\Handler\Locator\HandlerLocator;
use League\Tactician\Handler\MethodNameInflector\HandleInflector;
use League\Tactician\Handler\MethodNameInflector\MethodNameInflector;
use League\Tactician\Middleware;

class Builder
{
    private $middlewares = [];
    private $handlerLocator;
    private $commandNameExtractor;
    private $inflector;

    public function __construct(
        HandlerLocator $handlerLocator,
        CommandNameExtractor $commandNameExtractor = null,
        MethodNameInflector $inflector = null
    ) {
        $this->handlerLocator = $handlerLocator;
        $this->commandNameExtractor = $commandNameExtractor ?: new ClassNameExtractor();
        $this->inflector = $inflector ?: new HandleInflector();
    }

    public function attachMiddleware(Middleware $middleware, $priority = null)
    {
        if ($priority !== null) {
            $this->middlewares[$priority] = $middleware;
        } else {
            $this->middlewares[] = $middleware;
        }
    }

    public function build()
    {
        ksort($this->middlewares);
        $this->middlewares[] = new CommandHandlerMiddleware(
            $this->commandNameExtractor,
            $this->handlerLocator,
            $this->inflector
        );

        return new CommandBus($this->middlewares);
    }
}

```

And then, we can configure CommandBus using a configuration file, for example:

```php
<?php

$builder = new Builder(new ContainerBasedLocator($container));
$middlewares = require './middlewares.php';
foreach ($middlewares as $middleware) {
    if (is_array($middleware)) {
        if (sizeof($middleware) != 2) {
            throw new \LogicException('Invalid middleware definition.');
        }
        list($middleware, $priority) = $middleware;
    } else {
        $priority = null;
    }
    $builder->attachMiddleware($container->get($middleware), $priority);
}
$bus = $builder->build();
```

```php
<?php

// middlewares.php

return [
    [LockingMiddleware::class, -255],
    [LoggingMiddleware::class, 255],
    [TransactionMiddleware::class, 0],
];

```

rosstuck:
[...]

PR #61, opened 2015-04-16
Added CommandBus builder

Kilte:
See #58

### DIFF
--- a/src/Setup/Builder.php
+++ b/src/Setup/Builder.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace League\Tactician\Setup;
+
+use League\Tactician\CommandBus;
+use League\Tactician\Handler\CommandHandlerMiddleware;
+use League\Tactician\Handler\CommandNameExtractor\ClassNameExtractor;
+use League\Tactician\Handler\CommandNameExtractor\CommandNameExtractor;
+use League\Tactician\Handler\Locator\HandlerLocator;
+use League\Tactician\Handler\MethodNameInflector\HandleInflector;
+use League\Tactician\Handler\MethodNameInflector\MethodNameInflector;
+use League\Tactician\Middleware;
+
+class Builder
+{
+    /**
+     * @var Middleware[]
+     */
+    private $middlewares = [];
+
+    /**
+     * @var CommandHandlerMiddleware
+     */
+    private $commandHandler;
+
+    /**
+     * @param HandlerLocator       $locator
+     * @param CommandNameExtractor $extractor
+     * @param MethodNameInflector  $inflector
+     */
+    public function __construct(HandlerLocator $locator, CommandNameExtractor $extractor = null, MethodNameInflector $inflector = null)
+    {
+        $this->commandHandler = new CommandHandlerMiddleware(
+            $extractor ?: new ClassNameExtractor(),
+            $locator,
+            $inflector ?: new HandleInflector()
+        );
+    }
+
+    /**
+     * Pushes a middleware.
+     *
+     * @param Middleware $middleware
+     *
+     * @return $this
+     */
+    public function push(Middleware $middleware)
+    {
+        $this->middlewares[] = $middleware;
+
+        return $this;
+    }
+
+    /**
+     * Prepends a middleware.
+     *
+     * @param Middleware $middleware
+     *
+     * @return $this
+     */
+    public function prepend(Middleware $middleware)
+    {
+        array_unshift($this->middlewares, $middleware);
+
+        return $this;
+    }
+
+    /**
+     * Creates a CommandBus.
+     *
+     * @return CommandBus
+     */
+    public function build()
+    {
+        $middlewares = $this->middlewares;
+        $middlewares[] = $this->commandHandler;
+
+        return new CommandBus($middlewares);
+    }
+}
--- a/src/Setup/Builder.php
+++ b/src/Setup/Builder.php
@@ -47,8 +47,6 @@ class Builder
      */
     public function build()
     {
-        $middlewares = $this->middlewares;
-
-        return new CommandBus($middlewares);
+        return new CommandBus($this->middlewares);
     }
 }
--- a/src/Setup/Builder.php
+++ b/src/Setup/Builder.php
@@ -3,12 +3,6 @@
 namespace League\Tactician\Setup;
 
 use League\Tactician\CommandBus;
-use League\Tactician\Handler\CommandHandlerMiddleware;
-use League\Tactician\Handler\CommandNameExtractor\ClassNameExtractor;
-use League\Tactician\Handler\CommandNameExtractor\CommandNameExtractor;
-use League\Tactician\Handler\Locator\HandlerLocator;
-use League\Tactician\Handler\MethodNameInflector\HandleInflector;
-use League\Tactician\Handler\MethodNameInflector\MethodNameInflector;
 use League\Tactician\Middleware;
 
 class Builder
@@ -17,28 +11,6 @@ class Builder
      * @var Middleware[]
      */
     private $middlewares = [];
-
-    /**
-     * @var CommandHandlerMiddleware
-     */
-    private $commandHandler;
-
-    /**
-     * @param HandlerLocator       $locator
-     * @param CommandNameExtractor $extractor
-     * @param MethodNameInflector  $inflector
-     */
-    public function __construct(
-        HandlerLocator $locator,
-        CommandNameExtractor $extractor = null,
-        MethodNameInflector $inflector = null
-    ) {
-        $this->commandHandler = new CommandHandlerMiddleware(
-            $extractor ?: new ClassNameExtractor(),
-            $locator,
-            $inflector ?: new HandleInflector()
-        );
-    }
 
     /**
      * Pushes a middleware.
@@ -76,7 +48,6 @@ class Builder
     public function build()
     {
         $middlewares = $this->middlewares;
-        $middlewares[] = $this->commandHandler;
 
         return new CommandBus($middlewares);
     }
--- a/src/Setup/Builder.php
+++ b/src/Setup/Builder.php
@@ -28,8 +28,11 @@ class Builder
      * @param CommandNameExtractor $extractor
      * @param MethodNameInflector  $inflector
      */
-    public function __construct(HandlerLocator $locator, CommandNameExtractor $extractor = null, MethodNameInflector $inflector = null)
-    {
+    public function __construct(
+        HandlerLocator $locator,
+        CommandNameExtractor $extractor = null,
+        MethodNameInflector $inflector = null
+    ) {
         $this->commandHandler = new CommandHandlerMiddleware(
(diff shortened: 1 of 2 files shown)


sagikazarmark:
I am not so sure about the initial command handler addition. Why is it there?

Kilte:
@sagikazarmark, I think, in most cases command handler is necessary. Why not?

sagikazarmark:
No, it is not. It is a middleware. From a builder point of view it does not matter what middleware you add.

Kilte:
Hmm, okay. I will remove it.

Kilte:
IMO, for now builder has no sense. I would like to add priorities.

sagikazarmark:
No sense? Why did you propose it then?

Priorities? FIFO See stackphp's middleware builder.

Kilte:
This is my use case:

```php
abstract class ServiceProvider
{
    protected $subscribers = [];
    protected $middlewares = [];

    final public function getSubscribers()
    {
        if (!is_array($this->subscribers)) {
            throw new \LogicException(sprintf('Expects array, %s given.', gettype($this->subscribers)));
        }

        return $this->subscribers;
    }

    final public function getMiddlewares()
    {
        if (!is_array($this->middlewares)) {
            throw new \LogicException(sprintf('Expects array, %s given.', gettype($this->middlewares)));
        }

        return $this->middlewares;
    }

    abstract public function getDefinitionSource(array $config);
}

```

```php
use League\Tactician\CommandBus;
use League\Tactician\Handler\CommandHandlerMiddleware;
use League\Tactician\Handler\CommandNameExtractor\ClassNameExtractor;
use League\Tactician\Handler\CommandNameExtractor\CommandNameExtractor as CommandNameExtractorInterface;
use League\Tactician\Handler\Locator\HandlerLocator as HandlerLocatorInterface;
use League\Tactician\Handler\MethodNameInflector\HandleInflector;
use League\Tactician\Handler\MethodNameInflector\MethodNameInflector as MethodNameInflectorInterface;
use League\Tactician\Middleware as MiddlewareInterface;

class Builder
{
    private $middlewares = [];
    private $handlerLocator;
    private $commandNameExtractor;
    private $inflector;

    public function __construct(
        HandlerLocatorInterface $handlerLocator,
        CommandNameExtractorInterface $commandNameExtractor = null,
        MethodNameInflectorInterface $inflector = null
    ) {
        $this->handlerLocator = $handlerLocator;
        $this->commandNameExtractor = $commandNameExtractor ?: new ClassNameExtractor();
        $this->inflector = $inflector ?: new HandleInflector();
    }

    public function attachMiddleware(MiddlewareInterface $middleware, $priority = null)
    {
        if ($priority !== null) {
            $this->middlewares[$priority] = $middleware;
        } else {
            $this->middlewares[] = $middleware;
        }
    }

    public function build()
    {
        ksort($this->middlewares);
        $this->middlewares[] = new CommandHandlerMiddleware(
            $this->commandNameExtractor,
            $this->handlerLocator,
            $this->inflector
        );

        return new CommandBus($this->middlewares);
    }
}

```

```php
$container = $this->getContainer();
$builder = new CommandBus\Builder(new CommandBus\ContainerBasedLocator($container));

foreach ($this->providers as $provider) {
    if ($provider instanceof Provider\ServiceProvider) {
        foreach ($provider->getMiddlewares() as $middleware) {
            if (is_array($middleware)) {
                if (sizeof($middleware) != 2) {
                    throw new \LogicException('Invalid middleware definition.');
                }
                list($middleware, $priority) = $middleware;
            } else {
                $priority = null;
            }
            $builder->attachMiddleware($container->get($middleware), $priority);
        }
    }
}
return $this->commandBus = $builder->build();
```

And now I can write something like this:

```php
class Provider extends ServiceProvider
{
    protected $middlewares = [Middleware::class];
    public function getDefinitionSource(array $config)
    {
        $tokenHeader = array_key_exists('authentication.token_header', $config)
            ? $config['authentication.token_header']
            : 'X-Auth-Token';

        return [
            TokenStorageInterface::class => \DI\object(InMemoryTokenStorage::class),
            TokenResolverInterface::class => \DI\object(HeaderBasedTokenResolver::class)->constructor($tokenHeader),
            Middleware::class => \DI\object(Middleware::class)->constructor(
                \DI\link(TokenResolverInterface::class),
                \DI\link(PolicyInterface::class),
                \DI\link(TokenStorageInterface::class),
                \DI\link(LoggerInterface::class)
            ),
        ];
    }
}

```

Now let's imagine that we need to register `Provider2` after the `Provider1`. Doesn't matter, why. 
`Provider1` contains `Middleware1`.
`Provider2` contains `Middleware2`.
We need to register `Middleware2` before `Middleware1`.
Without priorities it is impossible.

For example:

```
class Provider1 extends ServiceProvider
{
    protected $middlewares = [Middleware1::class];
}
class Provider2 extends ServiceProvider
{
    protected $middlewares = [Middleware2::class];
}

// requirements for providers are satisfied, but not for middlewares
$providers = [
    Provider1::class,
    Provider2::class
];
```

```
class Provider1 extends ServiceProvider
{
    protected $middlewares = [Middleware1::class, 1];
}
class Provider2 extends ServiceProvider
{
    protected $middlewares = [Middleware2::class, -1];
}

// All requirements are satisfied
$providers = [
    Provider1::class,
    Provider2::class
];
```

@sagikazarmark what is your use case?

sagikazarmark:
Actually priority is not my first concern. That said, I am not against it in the builder as it will be strict in the command bus.

By design middlewares does not know about each other (except having the next callable in the chain controlled). No knowledge about the order. That said, it can make sense in the builder. If you insist on having a "main" middleware, I suggest you doing something like the following:

``` php
public function build()
{
	$middlewares = array_merge(
		$this->prependMiddlewares,
		[$this->mainMiddleware],
		$this->appendedMiddlewares
	);

	return new CommandBus($middlewares);
}
```

Kilte:
If we do not need priorities and the CommadHandlerMiddleware, then creating CommandBus using Builder makes no sense, because Builder can be replaced by simple array.

@sagikazarmark, what did you mean, when you said
>I have a few use cases. Actually this might be a big help in DI.

?

sagikazarmark:
Ah, I see what you call a Builder now. Actually, it is rather a factory, which might not fit into the core because of the following reasons:

- There is no really-really general implementation
- Since it is not general (but tied to your needs) there isn't really more than two or three use cases


What I call a builder is a simple container which accepts several middlewares and creates a command bus when build is called. The reason behind this is during service construction, it is harder to inject an array of dependencies into the command bus than creating it in a builder. The builder accepts several, but not an array of middlewares (through setter injection?). You can extend it before resolution. Since the CommandBus itself is considered to be final, it is not a problem to create a dependency between the Builder and the CommandBus.

Kilte:
Okay. Now everything is clear to me. 

Will this get merged? If so, let me to squash commits before.

rosstuck:
Hi fellas, I'll be out of the country for a few days so I'll review and comment on this ASAP but that'll likely be mid next week. Really sorry for the delay but I appreciate the effort here. :smile:

sagikazarmark:
I am actually against this approach as it is not covering many use cases generally.

Kilte:
Well, then I close it.

sagikazarmark:
@Kilte wait, I was not aware of the changes you made. I recommend you reopening the PR and let @rosstuck express his opinion. After reviewing the changes, I imagined something similar actually.


``` php
class BusBuilder
{
    /**
     * @var Middleware[]
     */
    private $middlewares = [];

    /**
     * Counting identifier used for building around a specific point
     *
     * @var integer
     */
    private $currentBase = -1;

    /**
     * @param Middleware $middleware
     */
    public function prepend(Middleware $middleware)
    {
        array_unshift($this->middlewares, $middleware);
        $this->currentBase++;
    }

    /**
     * @param Middleware $middleware
     */
    public function push(Middleware $middleware)
    {
        $this->middlewares[] = $middleware;
    }

    /**
     * Build the middleware stack around a "main" middleware
     *
     * @param Middleware $middleware
     *
     * @return CommandBus
     */
    public function buildAround(Middleware $middleware)
    {
        $middlewares = $this->middlewares;

        // To build the middlewares around a "main" middleware
        // the proper position in the array must be determined
        if ($this->currentBase === -1) {
            array_unshift($middlewares, $middleware);
        } else {
            array_splice($middlewares, $this->currentBase + 1, 0, $middleware);
        }

        return new CommandBus($middlewares);
    }

    /**
     * Build the middleware stack
     *
     * @return CommandBus
     */
    public function build()
    {
        return new CommandBus($this->middlewares);
    }
}
```

I think this implementation lets you create both a linear stack and a stack built around a specific middleware.

@Kilte what do you think?

Kilte:
Can we merge `build` and `buildAround` methods?

```php
public function build(Middleware $mainMiddleware = null)
{
    $middlewares = $this->middlewares;
    if ($mainMiddleware !== null) {
        if ($this->currentBase === -1) {
           array_unshift($middlewares, $mainMiddleware);
        } else {
            array_splice($middlewares, $this->currentBase + 1, 0,  $mainMiddleware);
        }
    }

    return new CommandBus($middlewares);
}
```

sagikazarmark:
I think it isn't really necessary. Why would you do that?

Kilte:
Hmm. Yes. In this case, it's not really necessary. 
But if we want to specify a main middleware before creating a bus, then ...

```php
public function split(Middleware $middleware) 
{
    $this->mainMiddleware = $middleware;
}

public function build()
{
    $middlewares = $this->middlewares;
    if ($this->mainMiddleware !== null) {
        if ($this->currentBase === -1) {
           array_unshift($middlewares, $this->mainMiddleware);
        } else {
            array_splice($middlewares, $this->currentBase + 1, 0,  $this->mainMiddleware);
        }
    }

    return new CommandBus($middlewares);
}
```

sagikazarmark:
The point is that you want to build a commandbus from a set of middlewares OR around a dedicated middleware. From a DI point of view, it doesn't matter if you pass the middleware before or after calling buildAround.

From a reusable point of view: if you pass the middleware in before (and store it) you are not able to build around another middleware anymore.

If you want your middleware in the stack, push it. If you have a dedicated middleware, build your stack around it without modifying the builder's state.

So I would say it is better to make the dedicated middleware a dependency of the building procedure.

rosstuck:
After reading over this one and talking to some folks working on framework integration pieces, I don't quite have enough of a use case in core to merge this one. :confused: When the topic originally came up, I thought it would be more of a Builder extension to the QuickStart object with a fluent interface for common config. That's pure sugar coating though and as we add more framework integrations, it may become less relevant anyways.

This PR is a nice piece of code and I can see how this would be very useful in your own provider, I just don't have a use case for it in core at this time. I feel really bad saying this because I know you've put a lot of energy into @Kilte and I hope it doesn't sour you too much. If there's some other ideas about how this might look, I'm willing to discuss it further. I just don't see where to go as it stands now. Nonetheless, thank you very much for your time, effort and dedication.

Kilte:
Don't worry, @rosstuck, all good =). I think, I can close PR and issue.

PR #63, opened 2015-04-22
Fix variable name in League\Container plugin doc

lku:


rosstuck:
Another great doc catch, thanks a ton!

PR #64, opened 2015-04-22
Fix link to LockingMiddleware plugin from Middleware page

lku:


rosstuck:
Ah, great catch, thanks a ton! :)

Issue #55, opened 2015-04-06
Cascading Locator


rizqidjamaluddin:
Would there be interest in a locator that iteratively attempts to use a stack of other locators, continuing to the next when the first one fails? I've been using this a lot lately, generally to have the option of assigning explicit locators when necessary and a sensible default otherwise. Example snippet:

``` php
class CascadingLocator implements HandlerLocator {

    /** @var HandleLocator[] */
    protected $locators;

    public function __construct (array $locators) {
        array_walk($locators, function($i){ if(!$i instanceof HandlerLocator) throw new NotALocatorException; });
        $this->locators = $locators;
    }

    public function getHandlerForCommand (Command $command) {
        foreach ($this->locators as $locator) {
            try {
                $handler = $locator->getHandlerForCommand($command);
            } catch (MissingHandlerException $e) {
                continue;
            }
            return $handler;
        }
        throw MissingHandlerException::forCommand($command);
    }
}

```

And thus usage would be wrapped:

``` php
$explicitBindings = new InMemoryLocator;
$defaultNameTransposingLocator = new SomeCustomTransposingLocator;
new CommandHandlerMiddleware($extractor, new CascadingLocator([$explicitBindings, $defaultNameTransposingLocator]), $inflector);
```

A proper PR would, naturally, need tests, hence this just being an Issue for now. It's really just a utility class, so I'm not sure if it deserves its own separate repository.

rosstuck:
Hi @rizqidjamaluddin! The idea of a Composite or Chain locator has come up a few times already, so it's safe to say there's interest.

I think the big question is should we add a `hasHandler()` method to the `HandlerLocator` interface, rather than using the exception as flow control? The recent split of the NameExtractor interface has also changed the semantics of this interface a bit since it now takes a string name instead of the actual command. I suspect that makes it a bit simpler to do here, but we'd need to update the Symfony and Silex providers though that shouldn't be too much work.

If you're interested in doing the `hasHandler()` refactoring, I'd be happy to see the PR and ensure the framework integrations get updated. Likewise, if anyone else has an alternative suggestions or better ideas, please pitch in. :smile: 

If we did add this, it doesn't have any deps so I agree with just shipping it in the core package as a regular ol' HandlerLocator, no Plugins namespace necessary.

rosstuck:
If you're not interested in doing the refactoring, btw, I can pick it up myself then swing back around to this but it'll be a couple of weeks probably.

rtuin:
[...]

PR #66, opened 2015-06-01
Add Locator to retrieve Handlers from a Container

mvriel:
In this PR I have added a Locator that is capable of retrieving
a handler from a Container implementing the Container Interop
standard / shared interface.

The Locator's default behaviour is to append the suffix 'Handler' to
the given Command Name (assuming the Command Name is an FQCN)
and uses that to find a handler in the Container that is provided
in the constructor.

A secondary behaviour for the Locator is that you can pass a map with
which you can map command names to specific identifiers with which
a handler can be returned. This is especially convenient when you are
using a container that operates based on identifiers and not class
names.

This specific locator depends on the `container-interop/container-interop`
package but I have explicitly mentioned this in the suggest section
of the composer because it is not a hard requirement for the usage of
Tactician, only when you want to use this new Locator.

Using unit tests and by manually performing tests using dummy data
I have been able to verify that no errors occur due to omitting this
dependency by default.

### DIFF
--- a/src/Handler/CommandNameExtractor/CommandHandlerNameExtractor.php
+++ b/src/Handler/CommandNameExtractor/CommandHandlerNameExtractor.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace League\Tactician\Handler\CommandNameExtractor;
+
+use League\Tactician\Exception\CanNotDetermineCommandNameException;
+
+/**
+ * Extract the name from a command so that the name can be determined
+ * by the context better than simply the class name
+ */
+interface CommandHandlerNameExtractor
+{
+    /**
+     * Extract the handler name from a command FQCN.
+     *
+     * @param string $command
+     *
+     * @return string
+     */
+    public function extract($command);
+}
--- a/src/Handler/CommandNameExtractor/IdentifierMapExtractor.php
+++ b/src/Handler/CommandNameExtractor/IdentifierMapExtractor.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace League\Tactician\Handler\CommandNameExtractor;
+
+/**
+ * Extract the an identifier based on the given map where the key of the map is the FQCN for the Command.
+ */
+class IdentifierMapExtractor implements CommandHandlerNameExtractor
+{
+    /**
+     * An identity map that maps the FQCN of a Command (as key) to an identifier (as value).
+     *
+     * @var string[]
+     */
+    private $identityMap = [];
+
+    /**
+     * Registers the identity map with this Extractor.
+     *
+     * @param string[] $identityMap
+     */
+    public function __construct(array $identityMap)
+    {
+        $this->identityMap = $identityMap;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function extract($command)
+    {
+        if (is_string($command) === false) {
+            throw new \InvalidArgumentException(
+                'A Command Handler Name Extractor expects the given command to be an FQCN'
+            );
+        }
+
+        if (isset($this->identityMap[$command]) === false) {
+            throw new \OutOfBoundsException('No identity known for Command of class "' . $command . '"');
+        }
+
+        return $this->identityMap[$command];
+    }
+}
--- a/src/Handler/CommandNameExtractor/SuffixedClassNameExtractor.php
+++ b/src/Handler/CommandNameExtractor/SuffixedClassNameExtractor.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace League\Tactician\Handler\CommandNameExtractor;
+
+/**
+ * Extract the name from the class with a suffix, defaults to 'Handler'.
+ */
+class SuffixedClassNameExtractor implements CommandHandlerNameExtractor
+{
+    /**
+     * @var string
+     */
+    private $suffix;
+
+    /**
+     * Determine the suffix to use.
+     *
+     * @param string $suffix
+     */
+    public function __construct($suffix = 'Handler')
+    {
+        $this->suffix = $suffix;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function extract($command)
+    {
+        if (is_string($command) === false) {
+            throw new \InvalidArgumentException(
+                'A Command Handler Name Extractor expects the given command to be an FQCN'
+            );
+        }
+
+        return $command . $this->suffix;
+    }
+}
--- a/src/Handler/Locator/ContainerLocator.php
+++ b/src/Handler/Locator/ContainerLocator.php
@@ -5,6 +5,7 @@ namespace League\Tactician\Handler\Locator;
 use Interop\Container\ContainerInterface;
 use League\Tactician\Exception\InvalidCommandException;
 use League\Tactician\Exception\MissingHandlerException;
+use League\Tactician\Handler\CommandNameExtractor\CommandHandlerNameExtractor;
 
 /**
  * Fetch handler instances from a Dependency Injection Container supporting the Interop specification.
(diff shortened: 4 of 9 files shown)


rosstuck:
This looks great! Normally, we'd split the container-interop part out to a separate plugin package but since it's just interfaces (and this is an oft-requested feature), I think the soft dep is pretty justifiable.

The one thing I think might be worth refactoring here is the behavior around command name. Technically, that ought to be in a `CommandNameExtractor` implementation, rather than a `CommandLocator`. The Handler suffix version is pretty clever and could be really useful to a lot of people, even those who don't want to use a container backend. Likewise, we could do the same for a CommandClassName => mapped string version (I'm genuinely surprised we didn't bother building that already. D'oh!)

So, if we split that behavior out into two different CommandNameExtractors, this would be perfect and useful to a broader audience; they'd just need to pick which one to use. If you're really counting on the fallback from one to another there, check out #55, we could do something similar there.

mvriel:
I am sorry but I am having trouble understanding which behaviour surrounding the command name that you'd like to extract to a CommandNameExtractor. Do you mean that a CommandNameExtractor should also be used for a CommandHandlerName?

To me the strategy determining the Command Handler's name is a responsibility of the Locator; or at least delegated by the Locator. This would mean that the constructor of the ContainerLoader would feature another dependency for the Command Handler Name Extractor.

And do I get it right that you identify the following patterns that you'd like to have extracted into CommandNameExtractors:

1. Suffix a string, defaults to 'Handler', to the given Command's FQCN
2. Given a lookup table to the CommandNameExtractor, try to map the FQCN of the Command to a value in that lookup table

The reason for the fallback behaviour is because I can imagine three scenarios:

1. The container uses identifiers to find the command handler with (the lookup)
2. The container uses handler classes that are not suffixed with 'Handler' but may even be in a different namespace (the lookup again)
3. The container uses a simplified lookup mechanism where 'Handler' is added as a suffix to the Command's FQCN (the Handler thing).

Since these two differ only slightly I thought I'd combine them into one as it feels like overkill to separate them. Though if we were to use CommandHandlerNameExtractors as described above then this could be used as a strategy and there is no need for separate versions

mvriel:
@rosstuck I have pushed a proof of concept based on your remarks; is this what you mean?
There is a distinct difference between a CommandNameExtractor and a CommandHandlerNameExtractor though; in the Locators we only know the FQCN of the Command but don't have the actual object, meaning that a Command Handler Name Extractor must accept a string.

A CommandNameExtractor however receives a Command object; meaning that these differ in what they accept.

mvriel:
Also: if this is what you mean I will fix the tests; didn't bother right now (also due to time ;) )

rosstuck:
In Tactician, we don't really have a separation between the CommandName and the HandlerName (there isn't really the concept of a HandlerName at all). The CommandNameExtractor just figures out what the identifier should be and the HandlerLocator returns the one for that identifier.

So, in this case, we wouldn't have a CommandHandlerNameExtractor, we'd just use the CommandNameExtractor. Because there's no other real use case for CommandNameExtractors anyways. 

Anyways, that'd be the quick solution for getting this merged. :)

On a future oriented note: That's the way it works now, since it lets folks change their naming strategy separate from their container, which was a real issue when we had the two coupled. I admit it's a bit suboptimal, this was a later refactoring and we probably need to address this better in the next release or 2.0. You can see the issues it's caused with Cascading Locators in #55.

At this point, my thought is making the HandlerLocator itself implement container-interop (been pondering this for a while and would solve a few issues) or passing the Extractor as a parameter somehow so the two can evolve separately.

If you've got any thoughts about that, lemme know. :)

mvriel:
@rosstuck There is just one problem with the CommandNameExtractor: The rest of the application passes a Command _object_ while the Locators pass a Command _FQCN_ (hence: string). This means that the current implementation of the Locators cannot consume normal CommandNameExtractors

rosstuck:
Sorry, to clarify: the locators aren't meant to purely pass an FQCN. They're meant to just pass any string name. Could be a Symfony service or tag name or the result of a `$command->getCommandName()` or whatever you'd like.

In the original version, the CommandLocators just took the entire command object, then implemented their own location logic. But that caused coupling issues when people had different styles of service naming and then had to reimplement the entire Locator just to change that part.

At the time, we also considered passing the Extractor as a dependency but for wiring together in frameworks, that meant either required constructor position (no go) or making it a parameter in the `$locator->getHandler` call, but then we might as well unwrap it (which is what we did)

mvriel:
@rosstuck the problem is not de output or its meaning of the Locator; the problem is the input of the CommandNameExtractor's `extract ` method.

The CommandNameExtractor interface requires that the `extract` method receives an object; but the Locator does not have an object.

rosstuck:
Indeed, I think we're in agreement but need to sync up a bit. I've got to start the work day here, maybe catch up tonight on IRC or Slack? :smile_cat:

mvriel:
I can be a bit busy this evening but I will definitely try to get a hold of you ;)

src/Handler/Locator/InMemoryLocator.php
```diff
@@ -15,7 +15,7 @@
  *      $inMemoryLocator->addHandler($myHandler, 'My\TaskAddedCommand');
  *
  *      // Returns $myHandler
- *      $inMemoryLocator->getHandlerForCommand(new My\TaskAddedCommand());
+ *      $inMemoryLocator->getHandlerForCommand('My\TaskAddedCommand');
```
malukenho: can be `My\TaskAddedCommand::class` ?

mvriel: That is an allowed notation; I have explicitly used the pre-php 5.5 notation because this library its minimum version is not 5.5

rosstuck: Naw, Tactician is only >= 5.5 :smile:

mvriel: Holy smokes! I have totally had that wrong in my head! :D

src/Handler/Locator/ContainerLocator.php
```diff
@@ -0,0 +1,127 @@
...
+     */
+    private function retrieveHandlerFromContainer($handlerIdentifier)
+    {
+        try {
+            $handler = $this->container->get($handlerIdentifier);
+            if (is_object($handler) === false) {
+                throw new \OutOfBoundsException();
+            }
+        } catch (\Exception $e) {
+            throw new \OutOfBoundsException();
```
renan: Is it worth to include the previous exception (`$e`) when throwing another one?

src/Handler/Locator/ContainerLocator.php
```diff
@@ -0,0 +1,127 @@
...
+     * @throws \OutOfBoundsException if the container does not return an object.
+     * @throws \OutOfBoundsException if the container throws an error when retrieving the handler.
+
+     * @return object
+     */
+    private function retrieveHandlerFromContainer($handlerIdentifier)
+    {
+        try {
+            $handler = $this->container->get($handlerIdentifier);
+            if (is_object($handler) === false) {
```
renan: This `if` block could be moved outside the try/catch block

src/Handler/Locator/InMemoryLocator.php
localheinz: @mvriel @rosstuck 

Given the time that has passed, and given that PHP 5.5 is not supported anymore, maybe we can update this?

Also see #102.

src/Handler/CommandNameExtractor/CommandHandlerNameExtractor.php
```diff
@@ -0,0 +1,21 @@
...
+ * Extract the name from a command so that the name can be determined
+ * by the context better than simply the class name
+ */
+interface CommandHandlerNameExtractor
+{
+    /**
+     * Extract the handler name from a command FQCN.
+     *
+     * @param string $command
+     *
```
localheinz: @mvriel 

Should we add

```php
/**
 * @throws \InvalidArgumentException
*/
```

here, since the implementation provided in this PR throws one?

src/Handler/CommandNameExtractor/IdentifierMapExtractor.php
```diff
@@ -0,0 +1,44 @@
...
+    private $identityMap = [];
+
+    /**
+     * Registers the identity map with this Extractor.
+     *
+     * @param string[] $identityMap
+     */
+    public function __construct(array $identityMap)
+    {
+        $this->identityMap = $identityMap;
```
localheinz: @mvriel 

Should we assert here that the keys are

* `string`s
* map to names of existing classes

?

src/Handler/CommandNameExtractor/SuffixedClassNameExtractor.php
```diff
@@ -0,0 +1,38 @@
...
+    private $suffix;
+
+    /**
+     * Determine the suffix to use.
+     *
+     * @param string $suffix
+     */
+    public function __construct($suffix = 'Handler')
+    {
+        $this->suffix = $suffix;
```
localheinz: @mvriel 

Should we assert here that `$suffix` is a `string`?

src/Handler/CommandNameExtractor/SuffixedClassNameExtractor.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function extract($command)
+    {
+        if (is_string($command) === false) {
```
localheinz: @mvriel 

Should we also assert here that `$command` is the name of an existing class?

src/Handler/Locator/ContainerLocator.php
```diff
@@ -0,0 +1,127 @@
...
+ *      // Or by using the addHandler method
+ *      $containerLocator->addHandler('My\TaskRemovedCommand', 'my.handler.removed.identifier');
+ *
+ *      // Returns an the instance belonging to the key 'my.handler.identifier' from the container.
+ *      $containerLocator->getHandlerForCommand('My\TaskAddedCommand');
+ */
+final class ContainerLocator implements HandlerLocator
+{
+    /** @var ContainerInterface */
+    private $container;
```
localheinz: @mvriel 

For consistency, should we turn this into a multi-line docblock, rather than single line comment

```php
/**
 * @var ContainerInterface
 */
```

?

rosstuck:
Closing this before @localheinz gets carried away :)

The container-interop work here is now fulfilled by the new version of https://github.com/thephpleague/tactician-container. The other part involving a command name extractor probably needs to be picked up separately. If this is something we'd like to carry forward with perhaps it's a good idea to restart this one from scratch sometime :)

localheinz:
@rosstuck 

Ha, nice! Thanks for the heads-up!

PR #78, opened 2015-06-25
Add bootstrap file for phpunit using composers autoloader

dennisdegreef:
Running PHPUnit fatals on missing files, since it has no autoloader registered from composer.

I added a bootstrap file to load that in.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -2,6 +2,7 @@
 
 <phpunit backupGlobals="false"
          backupStaticAttributes="false"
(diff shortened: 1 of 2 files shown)


rosstuck:
As discussed with @dennisdegreef in person, rejecting this one since it was caused by a misunderstanding with regards with the best way to run the tests.

Should be `./vendor/bin/phpunit` rather than using a globally installed one. Otherwise, it doesn't pick up the composer autoloading configuration. :smile:

PR #80, opened 2015-06-30
Fix incorrect example

adamlundrigan:
[CommandHandlerMiddleware](https://github.com/thephpleague/tactician/blob/master/src/Handler/CommandHandlerMiddleware.php#L36) takes a CommandNameExtractor concrete implementation as it's first argument


rosstuck:
Could've sworn I'd already merged this, really really sorry for the delay!

PR #83, opened 2015-07-28
Added function to modify the middleware chain

RonRademaker:
Hi,

I've added a function to the command bus to modify the middleware chain after creating it. Given the nature of the chain I don't see a way (yet) to add something to the end of the chain, only to the beginning. I think this gives users more flexibility in building the command bus and I prefer the syntax of a several method calls over passing everything in a constructor.

### DIFF
--- a/src/CommandBus.php
+++ b/src/CommandBus.php
@@ -45,7 +45,7 @@ class CommandBus
 
     /**
      * Put some middleware on the front of the middleware chain.
-     * 
+     *
      * @param \League\Tactician\Middleware $middleware
      */
     public function prependMiddlewareChain(Middleware $middleware)
--- a/src/CommandBus.php
+++ b/src/CommandBus.php
@@ -26,7 +26,7 @@ class CommandBus
     }
 
     /**
-     * Executes the given command and optionally returns a value
+     * Executes the given command and optionally returns a value.
      *
      * @param object $command
      *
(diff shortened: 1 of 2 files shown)


RonRademaker:
Can you see why scrutinizer failed? I don't see any error messages.

rosstuck:
Hi @RonRademaker, sorry for the delay answering your other PR, vacation got in the way. I'm going to sit down and go through all the open issues Friday, I promise. :)

My first reaction here is that I probably won't merge this as it stands right now. This has less to do with the PR and more to do with Tactician's current design: I want the objects to be immutable and adding a prepend middleware chain would violate that. What's under discussion now is extracting the middleware chain from the command bus lib altogether and make it its own immutable collection. That would probably allow prepending, appending, etc and give a lot more flexibility as a whole :smile: 

In the meantime, folks can still prepend by passing the array of middleware around as part of the build phase, rather than the entire command bus.

I can understand the stylistic preference of multiple method calls rather than using the constructor but using the constructor exclusively has several benefits, mainly that it's one time setup with no future worry about the state of the command bus. It also doesn't tempt people to modify it on the fly (very dangerous!) and means we only have one place to do chain construction/setup.

I'm not sure what's up with Scruitinizer at the moment but many kudos for sending such a well formed PR though, I wish everyone included tests!

RonRademaker:
Hi @rosstuck, hope you're having a good vacation.

I actually considered something like a making the chain an object, but that'd be a large change which also breaks the API so I decided against it. Still, I think it'd be a nice feature.

The main reason I'd like to be able to do this is because I'm creating an optional addon for our framework where I'd like users to influence the command bus after the addon created the default. So I'm actually looking for a way to do the setup in two places (one the default setup and one an optional addition to that setup). 

Am I correct in concluding that the reason you want the chain to be immutable is because otherwise it could lead to changing the bus on the fly causing commands to be treated differently (which would lead to unexpected results)? If so, I think I see an alternative solution. This would involve keeping the middleware chain array in the CommandBus and building the actual chain when the first command must be handled. This would make adding a few modification methods easy as all they do is some array manipulation. After creating the chain, the modification methods should throw exceptions indicating a change to the middleware chain is no longer allowed.

rosstuck:
Indeed, I think the Chain object makes sense. Frank de Jonge and I are looking how tactician and `league/pipeline` might work together on that. So, there'll definitely be more coming soon.

Regarding the immutable aspect, it's not just about the reconfiguring on the fly but as a general design principle as well. Even building the chain lazily would create potential error scenarios where you could get an exception under one situation but not another. The only way to prevent that is have the chain be immutable and only a complete chain can be passed into the command bus (if we even need a command bus then :wink:)

I'm not sure exactly what's going on in the bootstrap process to require this level of on-the-fly config but it might be better to elaborate or refocus on that side of it.

PR #92, opened 2015-11-19
Adding composer install instructions to readme

JesseObrien:
This is totally for my own benefit so when I'm forgetting what to type I can come here and copy/paste it to my terminal like a plebe. :)


afolson:
:+1:

PR #95, opened 2016-01-15
Added gitattributes file

hannesvdvreken:


boekkooi:
Why are you not exporting the `LICENSE`? and `tests`, `phpunit.xml.dist`?

hannesvdvreken:
Why would you need test files when all that is autoloaded when installed via composer is the src folder? License, I don't know, isn't it enough that it's defined in the composer.json file?

rosstuck:
Wow, I never got an email for the original PR so I totally missed this. Many apologies @hannesvdvreken!

The exports here are in-line with the league recommendations so I'm inclined to accept them as is. Thanks!

hannesvdvreken:
:+1: no problem, you're welcome!

PR #96, opened 2016-02-19
Exceptional exceptions

boekkooi:
Let's avoid the exceptions throw a exception....

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,6 +9,7 @@
          processIsolation="false"
          stopOnFailure="false"
          syntaxCheck="false"
+         bootstrap="vendor/autoload.php"
         >
 
     <listeners>
--- a/src/Exception/CanNotDetermineCommandNameException.php
+++ b/src/Exception/CanNotDetermineCommandNameException.php
@@ -8,18 +8,20 @@ namespace League\Tactician\Exception;
 class CanNotDetermineCommandNameException extends \RuntimeException implements Exception
 {
     /**
-     * @var object
+     * @var mixed
      */
     private $command;
 
     /**
-     * @param object $command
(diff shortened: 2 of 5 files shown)


RonRademaker:
You use non-objects as commands? What's the use case for that?

boekkooi:
@RonRademaker well generally no I don't I had a bug in one of my middleware's that triggered a `get_class() expects parameter 1 to be object, array given`  and not the expected exception.

Also it's just handy to have the bootstrap in there for people that have phpunit installed on there system, it avoid a failing test suite when they just type `phpunit`.

RonRademaker:
Ah, I get the improvement now. Perhaps nice to also try to drop a ```callable``` in there in the tests?

I don't agree on the ```phpunit``` front though. Because you can't be sure a correct ```phpunit``` has been installed globally, so your test results may be unexpected or unreliable. In such cases it'd be better IMO to fail quickly.

boekkooi:
Just added a callable.

@rosstuck please give us you phpunit expert opinion because I think we need to agree to disagree @RonRademaker.

rosstuck:
The extra check is a nice touch, thanks!

With regards to PHPUnit, @RonRademaker is correct but since the CI server is configured to use the vendor version, it'll fail once they push it so I don't mind. Technically though, it should've been a separate PR (finger wag).

If you can fix the CS violation that's currently breaking the build, I'll happily merge this and tag a new release. Thanks to you both!

boekkooi:
Fixed the codestyle issue which makes it less readable but ok..... 

Yeah yeah I know it should be 2 PR's Do you want me to split it up?

rosstuck:
Naw, it's fine. Thanks again!

PR #97, opened 2016-03-07
docs: `getHandlerForCommand` receives the command name, not the Command object

guiwoda:
Just a small change in the `InMemoryLocator` class docs.


rosstuck:
Great catch, thanks! :)

PR #101, opened 2016-07-23
Fix: Namespace does not match project structure

localheinz:
This PR

* [x] fixes namespaces of tests which do not match the project structure


rosstuck:
Sure, thanks! :smile:

localheinz:
Thank you, @rosstuck!

PR #102, opened 2016-07-23
Fix: Limit range of accepted PHP versions

localheinz:
This PR

* [x] limits the range of accepted PHP versions

💁 We can't tell for sure yet whether this package will be compatible with PHP 9000, can we?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "php":  ">=5.5"
+        "php":  "^5.5 || ^7.0"
     },
     "require-dev": {
         "mockery/mockery": "~0.9",


localheinz:
@rosstuck 

While we're at it, maybe worth dropping support for PHP 5.5?

GrahamCampbell:
The league's policy is not to drop for no reason.

GrahamCampbell:
If there are some PHP 5.6 features that we want to actually use, that's when 5.5 would be dropped.

rosstuck:
@localheinz Thanks but at this time, I'm okay with waiting to see what future versions will bring before limiting it to an upper bound. Tactician is relatively small and with PHP's general BC policy, I think we'll be okay with adding it to the test matrix when the time comes. With regards to 5.5, it's officially EOL. I may deprecate at a future point (2.0, if there's ever a need).

@GrahamCampbell I'm aware of League policy, it's a decision I'll make when relevant.

frankdejonge:
@rosstuck while it's common to do so, I don't believe there's actually a policy to prevent the deprecation of PHP version support. /cc @GrahamCampbell

GrahamCampbell:
Phil mentioned it a few months ago. Sorry I can't remember where. It was basically along the lines of don't do it for the sake of doing it. As in, there is no policy to force bump the league packages.

frankdejonge:
True, but that's an opinion, not a policy. I think a pretty good case can be made to up when a version is not supported anymore, if only to change/upgrade gradually without a massive jump from say 5.3 to 7.1.

PR #103, opened 2016-10-08
Fix bernard example

sagikazarmark:
@rosstuck Just realized this while preparing my Tactician talk

Although I am not 100% happy with this solution, it's okay temporarily

### DIFF
--- a/plugins/bernard.md
+++ b/plugins/bernard.md
@@ -51,7 +51,12 @@ $commandBus = new CommandBus([/* middlewares go here */]);
 // Wire the command bus into Bernard's routing system
 $receiver = new SingleBusReceiver($commandBus);
 $router = new SimpleRouter();
-$router->add('League\Tactician\Bernard\QueueableCommand', $receiver);
+
+foreach(get_declared_classes() as $class) {
+    if(is_a($class, 'League\Tactician\Bernard\QueueableCommand', true)) {
+        $router->add($class, $receiver);
+    }
+}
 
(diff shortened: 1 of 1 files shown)


rosstuck:
Hey, you're working on a Tactician talk? Cool, let me know if there's going to be a video or slides online, would love to see it :)

With this one, I totally see it as a quick fix but I'm worried about putting it in the docs. If there's one thing I've learned over the last couple years, it's that folks tend to take this stuff very literally and if we put this in there, I promise you more than a few well-meaning folks will actually put this in their applications and it'll stay there. How about we change it to adding a few "queue-able" commands as an example, then mention you can automate this with DI tags, naming conventions, config handlers, etc, whatever makes sense for people's apps.

Would that be a fair compromise? :)

sagikazarmark:
> Hey, you're working on a Tactician talk? Cool, let me know if there's going to be a video or slides online, would love to see it :)

Will do.

> Would that be a fair compromise? :)

Yeah, sounds OK. PR was a good idea though, at least I won't forget about it.

rosstuck:
:+1: Cool, sounds good! It is a clever trick, kudos :)

sagikazarmark:
Seems like class names don't work either. Bernard escapes the namespaces too. I wonder how I used this in production so far.

rosstuck:
Gonna close this one for now. I'm waiting to see how the naming convention stuff works in the bundle and then whatever ends up working there successfully will go into Tactician2 :)

PR #104, opened 2016-11-07
PHP 5.4 compatibility

bursonic:
Realized that main code doesn't use any features of 5.5+. 
Incompatibility is caused only by ::class magic constant usage in tests and examples.

So why not lower version according to real feature usage?

Unfortunately 5.4 is still widely set as a requirement: https://seld.be/notes/php-versions-stats-2016-1-edition
There are thousands of legacy projects (including mine) still using this version in production.
I know that 5.4 is an ancient thing now but this change could help other people with such real-life limitations to use the package without additional workarounds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.idea
 vendor/
 composer.lock
 phpunit.xml
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 vendor/
 composer.lock
 phpunit.xml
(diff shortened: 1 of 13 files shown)


rosstuck:
Hi @bursonic! Thanks for taking the time to send this PR. 

Unfortunately, 5.4 has been EOL'ed for quite some time and I would rather not support it, both for practical reasons (I quite like the ::class constant!) and also to set an example going forward. Going ahead with a version that doesn't have security updates is just super dangerous for you and your users.

I am really sorry that you've spent time on this only to have it rejected, my apologies for that and I hope it doesn't put you off. Please feel free to open an issue next time and we can discuss it beforehand. :)

PR #107, opened 2017-02-04
Add Tactician Domain Events plugin

borNfreee:
Previously discussed in https://github.com/thephpleague/tactician-doctrine/issues/15


rosstuck:
Thanks a ton! :)

PR #111, opened 2017-04-12
Introduce a PSR-11 handler locator

chalasr:
First thanks for this library, I use it in pretty much all of my personal/enterprise projects.

This allows to lazy load handlers from any PSR-11 container.
In addition to being useful on its own, the final target of adding this locator is to make your `tactician-bundle` leverage the changes made in Symfony 3.3: we added the ability to lazy load services through explicit service locators, allowing to stop injecting the whole DI container (that is just bad design, it hides dependencies). See https://github.com/symfony/symfony/pull/21553, https://github.com/symfony/symfony/pull/21708 for references.

If accepted I would be glad to make your bundle uses this through the symfony dependency-injection component, replacing the current ContainerBasedHandlerLocator which is coupled to the symfony DIC. The main advantage for end users being that command handler services could be made private, leveraging the optimisations that the symfony DI performs on private services (better performances, better design).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
     "require-dev": {
         "mockery/mockery": "~0.9",
         "phpunit/phpunit": "~4.3",
+        "psr/container": "^1.0",
         "squizlabs/php_codesniffer": "~2.3"
     },
     "autoload": {
--- a/src/Handler/Locator/ContainerLocator.php
+++ b/src/Handler/Locator/ContainerLocator.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace League\Tactician\Handler\Locator;
+
+use League\Tactician\Exception\MissingHandlerException;
+use Psr\Container\ContainerInterface;
+
+/**
+ * Lazily loads Command Handlers from a PSR-11 container.
+ */
+class ContainerLocator implements HandlerLocator
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    /**
+     * @param ContainerInterface $container A PSR-11 container holding command handlers
+     *                                      mapped by command names
+     */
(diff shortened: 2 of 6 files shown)


rosstuck:
Hi @chalasr! Thanks for this excellent looking PR, I really appreciate the work you've put into it. :)

Unfortunately, this functionality is already covered in https://github.com/thephpleague/tactician-container, rather than the main category. I've just merged an existing PR there this morning and tagged a new release. If you'd like to give it a review and any improvements, please do! 

I'm sorry that you've put time into this. In the future, feel free to ping me about an idea by opening an issue and I'll be happy to help coordinate stuff so there's no duplicate work.

That said, adding this to the bundle version sounds like a really good idea! I don't believe anyone is working on that so if you'd like to pick it up there, I'd suggest opening an issue and starting on a PR there. The only gotcha is that I don't have time right now to maintain the bundle (renovating a house and moving this month) and it has some outstanding issues. The most excellent Rafael Dohms and Frank de Jonge are helping out but I think we're all a bit snowed under so a review might take a little while. I apologize in advance for that but I think this would be a really good addition, so please go for it!

Thanks again for your hard work. 😸

rosstuck:
(just closing in favor of the league/container implementation. If I've missed anything, please let me know and I'll reopen)

chalasr:
@rosstuck Happy to see that there is already something in! Don't worry about the work made here, fortunately PSR-11 and your HandlerLocator interfaces are quite easy to implement. I should have looked a bit more at your repositories beforehand. 

Btw, what about adding `tactician-container` as a `suggest` in the composer.json here so that people already using this library can be aware of it on the next upgrade?

I'll prepare something leveraging the `tactician-container` lib in the bundle asap.
Thanks!

Issue #113, opened 2017-08-15
We created queue interop based middleware for tactician.


makasim:
It uses [queue interop](https://github.com/queue-interop/queue-interop) compatible transports to send and consume tactician commands. 

The code: https://github.com/formapro-forks/tactician-queue-interop
Usage: https://gist.github.com/ASKozienko/f85eee441f345d292775669c7d48f6b8

Would you be interested to see it in thephpleague org or we should simply submit a link to it?

rosstuck:
Hi @makasim! That sounds really cool, thanks for posting about it. 

At this time, I'm not planning on adding anymore top-level Tactician plugins to the league namespace. That said, if you'd like to open a PR to have it mentioned in the Community Plugin section of the docs (see gh-pages branch of this bundle), I'd be more than happy to merge it!

I'll close this issue for now in lieu of a docs PR but if you feel there's anything else you'd like to talk about, don't hesitate to reopen it. :)

Thanks again!

PR #114, opened 2017-08-16
add link to queue interop transport.

makasim:
as suggested in https://github.com/thephpleague/tactician/issues/113


rosstuck:
Done and done. Thanks again! :)

makasim:
Thank you too.

PR #115, opened 2017-08-16
Make private props protected for using it in inherited classes

vtvz:


snapshotpl:
Inheritance is evil

rosstuck:
@vtvz Thanks for sending your PR! After giving it some thought and considering how small this class is, I think it's better to not open it for extension. It's probably better if folks reimplement the middleware in a way that they prefer, especially since they can hardcore their own conventions, rather than deal with the all the dependencies like the extractor and stuff (which are only there for the sake of configuration, really).

So, with respect, I'm going to pass on this PR for now. If you feel there's a very important use case I've missed, please let me know and we can discuss reopening the ticket.

@snapshotpl Thanks for taking the time to peek at these PRs and comment on them. At the same time, if you'd like to leave feedback, please try to do so in a way that's constructive and helpful. Just saying something is good or evil without talking about why doesn't help folks improve or provide better discussion. I do appreciate wanting to contribute (lord knows I can use the help! 😄 ), but Tactician is just as much about helping folks feel good sending their first PR as it is about having the best code we can. :)

RonRademaker:
@vtvz Of course I don't know your usecase, but the middleware pattern used by tactician is very well suited to change how your commands are handled without having to extend anything. For example, if you added something to the `execute` method like:

``` php
public function execute($command, callable $next) 
{
    // do some stuff first
   return parent::execute($command, $next);
}
```

You can just put `// do some stuff` in some middleware of your own, like:

``` php
class MyMiddleware implements Middleware 
{
   public function execute($command, callable $next) 
   {
        // do some stuff first
       return $next($command);
   }
}
```

vtvz:
@rosstuck I wanted make my own implementation of `CommandHandlerMiddleware` which logic differs just a little bit. It passes command by setter (`setCommand`) and calls `$methodName` without `$command` param. For me is't better to extend your `CommandHandlerMiddleware` and reimplement `execute` method. But I cant use `commandNameExtractor`, `handlerLocator` and `methodNameInflector` in my code because this props are private. And I forced to copy and paste your code and write almost the same.
May be you can write getters like `getCommandNameExtractor` in `CommandHandlerMiddleware` to use it in nested classes? It will be perfect for me

rosstuck:
@vtvz In this case, I think it's probably better to copy and paste the code and just change/simplify it to your needs. You can still pass in instances of of ```CommandNameExtractor```, ```HandlerLocator``` and ```MethodInflector``` if you want, but if you're building for your use case, it's probably simpler to not use them. There's nothing wrong with getting your instances or copy&pasting this small amount of code. :)

PR #118, opened 2017-11-19
Class Name Inflector

tomaszhanc:
I would like to have handler methods named without `handle` prefix. For example: `rentMovie` instead of `handleRentMovie`.

PR for docs: https://github.com/thephpleague/tactician/pull/119


rosstuck:
Thanks!

For future readers: In theory, I'm not super keen on supporting every variation of this but this one seems pretty well assembled so I'll merge it but this is probably the last one. :)

PR #120, opened 2017-12-19
Fix LockingMiddleware - not wothking for PHP7

evangellion:
Add missing support for Throwable

### DIFF
--- a/src/Plugins/LockingMiddleware.php
+++ b/src/Plugins/LockingMiddleware.php
@@ -44,9 +44,9 @@ class LockingMiddleware implements Middleware
         try {
             $returnValue = $this->executeQueuedJobs();
         } catch (\Exception $e) {
-            $this->isExecuting = false;
-            $this->queue = [];
-            throw $e;
+            $this->unlockOnException($e);
+        } catch (\Throwable $e) {
+            $this->unlockOnException($e);
         }
 
         $this->isExecuting = false;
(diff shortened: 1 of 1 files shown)


rosstuck:
Hi @evangellion, thanks for opening this PR! The PR looks like it makes sense, though can you take a look at the build failures? I'm a bit short on time this week but I'll try to merge once it looks green everywhere. :)

src/Plugins/LockingMiddleware.php
```diff
@@ -68,4 +68,12 @@ protected function executeQueuedJobs()
 
         return array_shift($returnValues);
     }
+    
+    private function unlockOnException($e): void
```
Txumari: I think you can't use return type declarations for php5 compatibility

src/Plugins/LockingMiddleware.php
```diff
@@ -68,4 +68,12 @@ protected function executeQueuedJobs()
...
+    {
+        $this->isExecuting = false;
+        $this->queue = [];
+        throw $e;
+    }
+
```
Txumari: style fix: remove this white line

src/Plugins/LockingMiddleware.php
```diff
@@ -68,4 +68,12 @@ protected function executeQueuedJobs()
 
         return array_shift($returnValues);
     }
+    
+    private function unlockOnException($e): void
+    {
+        $this->isExecuting = false;
+        $this->queue = [];
+        throw $e;
```
ogizanagi: Detail, but: re-throwing should probably be done in the catch clauses directly in order to avoid cluttering the stack trace with a useless item.

PR #121, opened 2018-04-04
Add CakePHP framework integration plugin.

robertpustulka:
There's a plugin I've been working on for some time: https://github.com/robotusers/cakephp-tactician

Would be great if it appeared on the docs page and helped other CakePHP devs :smiley:


rosstuck:
Sure thing, thanks for supporting Tactician like this! :)

PR #125, opened 2018-08-21
Add wikipedia link to example

yourwebmaker:
Just adds the wikipedia link mentioned in the examples documentation.


rosstuck:
Good one, thanks!

PR #127, opened 2018-11-30
Drop EOLed PHP versions

simPod:
I saw this on github

image

so I was like, yea, let's do this! Let's use modern standards!

[PHP 7.1 support is gone when you wake up this morning](http://php.net/supported-versions.php)

image

It's been great ride but it's time to say goodbye.

Let's use all those new features PHP has to offer and don't get stuck on PHP legacy versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,18 @@ language: php
 dist: trusty
 
 php:
-    - 5.5
-    - 5.6
-    - 7.0
-    - 7.1
     - 7.2
+    - 7.3
+    - nightly
 
 matrix:
     include:
-        - php: 5.5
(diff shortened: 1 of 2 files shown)


rosstuck:
Aww yiss!

In all seriousness, I'm on vacation right now so this will have to wait 10
days or so, sorry. I'm very inclined to drop 7.0 support, may drop 7.1
support in next major version so we can get that sweet sweet object
typehint finally. :)

Thanks for the nudge on this!

rosstuck:
Just a quick update, I'm likely going to go through and review all the Tactician packages and update them accordingly. That'll be a bigger effort (since I want to change typehints, interfaces, etc) so this'll happen but when I have a bit more time to sink into everything at once. :)

rosstuck:
Happening now in master :)

PR #131, opened 2019-08-09
Fixed the cats ears in the example

vanbrabantf:
Currently, the sparkling eyes make it seem like the cat has lost his/her ears.
This PR tries to solve this by updating the eyes back to dots and add an "undertie" (U+203F) for a mouth.


vanbrabantf:
I hope it was ok to do this over PR due to the sensitivity of the topic.

PR #132, opened 2019-08-09
v2 improvements

snapshotpl:
* Make all implementation final,
* add typed args where possible,
* add suffix handler named constructor,
* support for php 7.4

### DIFF
--- a/src/Handler/CanNotInvokeHandler.php
+++ b/src/Handler/CanNotInvokeHandler.php
@@ -22,14 +22,11 @@ final class CanNotInvokeHandler extends BadMethodCallException implements Except
     /** @var object */
     private $command;
 
-    /**
-     * @return static
-     */
-    public static function forCommand(object $command, string $reason)
+    public static function forCommand(object $command, string $reason): self
     {
         $type = get_class($command);
 
-        $exception          = new static(
+        $exception = new self(
             'Could not invoke handler for command ' . $type .
             ' for reason: ' . $reason
         );
--- a/src/Handler/CanNotInvokeHandler.php
+++ b/src/Handler/CanNotInvokeHandler.php
@@ -6,6 +6,7 @@ namespace League\Tactician\Handler;
 
 use BadMethodCallException;
 use League\Tactician\Exception;
+use Throwable;
 use function get_class;
 use function gettype;
 use function is_object;
@@ -16,19 +17,17 @@ use function is_object;
  * The most common reason is the receiving method is missing or incorrectly
  * named.
  */
-class CanNotInvokeHandler extends BadMethodCallException implements Exception
+final class CanNotInvokeHandler extends BadMethodCallException implements Exception
 {
-    /** @var mixed */
+    /** @var object */
     private $command;
 
     /**
-     * @param mixed $command
-     *
      * @return static
      */
-    public static function forCommand($command, string $reason)
+    public static function forCommand(object $command, string $reason)
     {
-        $type =  is_object($command) ? get_class($command) : gettype($command);
+        $type = get_class($command);
 
         $exception          = new static(
             'Could not invoke handler for command ' . $type .
@@ -39,12 +38,15 @@ class CanNotInvokeHandler extends BadMethodCallException implements Exception
         return $exception;
     }
 
+    private function __construct(string $message = '', $code = 0, Throwable $previous = null)
+    {
+        parent::__construct($message, $code, $previous);
+    }
+
     /**
      * Returns the command that could not be invoked
-     *
-     * @return mixed
      */
-    public function getCommand()
+    public function getCommand(): object
     {
         return $this->command;
     }
--- a/src/Handler/CommandHandlerMiddleware.php
+++ b/src/Handler/CommandHandlerMiddleware.php
@@ -13,7 +13,7 @@ use function is_callable;
 /**
  * The "core" CommandBus. Locates the appropriate handler and executes command.
  */
-class CommandHandlerMiddleware implements Middleware
+final class CommandHandlerMiddleware implements Middleware
 {
     /** @var ContainerInterface */
     private $container;
--- a/src/Handler/Mapping/ClassName/Suffix.php
+++ b/src/Handler/Mapping/ClassName/Suffix.php
@@ -4,10 +4,17 @@ declare(strict_types=1);
 
 namespace League\Tactician\Handler\Mapping\ClassName;
 
-class Suffix implements ClassNameInflector
+final class Suffix implements ClassNameInflector
 {
+    private const DEFAULT = 'Handler';
+
     /** @var string */
     private $suffix;
+
+    public static function handler(): self
+    {
+        return new self(self::DEFAULT);
+    }
 
     public function __construct(string $suffix)
     {
--- a/src/Handler/Mapping/ClassName/Suffix.php
+++ b/src/Handler/Mapping/ClassName/Suffix.php
@@ -6,15 +6,8 @@ namespace League\Tactician\Handler\Mapping\ClassName;
 
 final class Suffix implements ClassNameInflector
 {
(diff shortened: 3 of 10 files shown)


src/Handler/CanNotInvokeHandler.php
```diff
@@ -39,12 +35,15 @@ public static function forCommand($command, string $reason)
         return $exception;
     }
 
+    private function __construct(string $message = '', $code = 0, Throwable $previous = null)
+    {
+        parent::__construct($message, $code, $previous);
+    }
+
```
rosstuck: I like the intent here but I prefer to leave the constructor open so other folks can potentially use it down stream. Maybe they'd like to format something entirely different here but still use a Tactician level exception. :)

src/Handler/CanNotInvokeHandler.php
```diff
@@ -39,12 +35,15 @@ public static function forCommand($command, string $reason)
...
     /**
      * Returns the command that could not be invoked
-     *
-     * @return mixed
      */
-    public function getCommand()
+    public function getCommand(): object
     {
         return $this->command;
     }
```
rosstuck: Other than that one thing, I love everything you've done in this file! :heart: :rainbow: :unicorn: :star2:

src/Handler/CommandHandlerMiddleware.php
```diff
@@ -13,7 +13,7 @@
 /**
  * The "core" CommandBus. Locates the appropriate handler and executes command.
  */
-class CommandHandlerMiddleware implements Middleware
+final class CommandHandlerMiddleware implements Middleware
 {
     /** @var ContainerInterface */
```
rosstuck: Groovy! I need to reset my IDE config, le sigh.

src/Handler/Mapping/ClassName/Suffix.php
```diff
@@ -4,11 +4,18 @@
...
+    private const DEFAULT = 'Handler';
+
     /** @var string */
     private $suffix;
 
+    public static function handler(): self
+    {
+        return new self(self::DEFAULT);
+    }
+
```
rosstuck: Dig the final and while I like the idea, I'd like to avoid supporting too many direct defaults in T2, so would prefer to drop this.

src/Handler/Mapping/MethodName/Handle.php
```diff
@@ -7,7 +7,7 @@
 /**
  * Handle command by calling the "handle" method.
  */
-class Handle implements MethodNameInflector
+final class Handle implements MethodNameInflector
 {
     /**
      * {@inheritdoc}
```
rosstuck: Also groovy! I should just setup my own coding standards thing, similar to doctrine-coding-standards

src/Handler/Mapping/MethodName/HandleClassNameWithoutSuffix.php
```diff
@@ -34,11 +36,12 @@ public function __construct(string $suffix = 'Command')
     {
         $this->suffix       = $suffix;
         $this->suffixLength = strlen($suffix);
+        $this->handleLastPartOfClassName = new HandleLastPartOfClassName();
     }
 
     public function getMethodName(string $commandClassName) : string
     {
-        $methodName = parent::getMethodName($commandClassName);
+        $methodName = $this->handleLastPartOfClassName->getMethodName($commandClassName);
```
rosstuck: I normally avoid constructing anything in the constructor but in this case I'll allow it. :ok_woman: :heavy_check_mark:

src/Handler/Mapping/MethodName/HandleLastPartOfClassName.php
```diff
@@ -13,14 +13,22 @@
...
+final class HandleLastPartOfClassName implements MethodNameInflector
 {
+    /** @var LastPartOfClassName */
+    private $lastPartOfClassName;
+
+    public function __construct()
+    {
+        $this->lastPartOfClassName = new LastPartOfClassName();
+    }
+
```
rosstuck: Same as above :)

src/Handler/Mapping/MethodName/Invoke.php
```diff
@@ -8,7 +8,7 @@
  * Handle command by calling the __invoke magic method. Handy for single
  * use classes or closures.
  */
-class Invoke implements MethodNameInflector
+final class Invoke implements MethodNameInflector
```
rosstuck: Triple groovy! I should buy a boat.

src/Handler/Mapping/MethodName/LastPartOfClassName.php
```diff
@@ -16,7 +16,7 @@
  *  - \MyGlobalCommand    => $handler->myGlobalCommand()
  *  - \My\App\CreateUser  => $handler->createUser()
  */
-class LastPartOfClassName implements MethodNameInflector
+final class LastPartOfClassName implements MethodNameInflector
```
rosstuck: Quad groo-, you know what, you get the idea

tests/Handler/Mapping/ClassName/SuffixTest.php
```diff
@@ -12,12 +12,19 @@ class SuffixTest extends TestCase
...
         );
     }
 
+    public function testBuildHandlerSuffix(): void
+    {
+        $result = Suffix::handler()->getClassName('MyCommand');
+
+        self::assertSame('MyCommandHandler', $result);
+    }
+
```
rosstuck: In line with previous comment about ::handle this would go too, sorry :crying_cat_face:

tests/Handler/Mapping/ClassName/SuffixTest.php
```diff
@@ -12,12 +12,19 @@ class SuffixTest extends TestCase
     /** @dataProvider examples */
     public function testCanAddSuffixToClassName(string $suffix, string $command, string $expectedResult) : void
     {
-        self::assertEquals(
+        self::assertSame(
```
rosstuck: No complaints but curious why the change, just general strictness? :smile_cat:

snapshotpl: Strictness is safely :)

src/Handler/CanNotInvokeHandler.php
snapshotpl: Look at `command` property. And getter. After one constructor instance will be broken.

rosstuck: Yeah, there is no way to set it short of doing something dumb like a protected setter. Alright, cool, let's keep it :)

tests/Handler/Mapping/ClassName/SuffixTest.php
rosstuck: Or the illusion of safety sometimes, but duly noted :+1:

rosstuck:
Thanks a lot for all your work on this! I really appreciate it :)

snapshotpl:
Thanks for your review! BTW what are you think about make `handle(): void` in `CommandBus`?

rosstuck:
Tactician explicitly allows return types (because synchronous commands aren't a bad thing), so `void` return type is a no-go. For extra safety, I've been working on phpstan-tactician which supports checking this, even if PHP's type system doesn't support it.

snapshotpl:
@rosstuck do you need some help with v2 release?

rosstuck:
Thanks for asking! The Tactician-PHPStan stuff is working so I'm mainly punching through the existing packages and updating things. Tactician-Logger and Tactician-Doctrine are updated, I'll deprecate Bernard and Command-Events. The bundle is the big challenge, I'm investigating options for that now. After that, need to update all the new docs and then tag all the releases. If the bundle work takes too long, I'll just tag the rest without but would prefer to do things in one fell swoop.

Not sure if any of that sounds interesting but feel free to let me know if something does. :)

PR #133, opened 2019-08-11
Make marker interface extend throwable

BackEndTea:
This makes sure the interface can only be used by exceptions/errors, and it means that tooling can understand that this can be 'caught.'


rosstuck:
Sorry for missing this one, good change though!

BackEndTea:
Thank you @rosstuck

PR #135, opened 2019-10-02
Remove unused imports

mark-gerarts:
While browsing the repository I noticed some unused imports, which appear to be introduced in https://github.com/thephpleague/tactician/commit/25b977e8e07332c3d1e9f21796e284afbb4356c2. After quick inspection it seems they were also unused in a related test.


rosstuck:
Yesssss, excellent catch

yesss

PR #137, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix ./ --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/examples/2-intermediate-create-middleware.php
+++ b/examples/2-intermediate-create-middleware.php
@@ -38,7 +38,7 @@ class LoggingMiddleware implements Middleware
 
     public function execute($command, callable $next)
     {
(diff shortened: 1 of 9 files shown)


rosstuck:
I'll move these over to `use function` import calls instead but this is still a nice improvement. Thanks a lot!

GrahamCampbell:
That blog post talks about only importing those that are actually optimised. Not all functions from the global namespace are optimized by the compiler.

PR #139, opened 2020-02-12
Construcor parameter type of CanNotInvokeHandler

szepeviktor:
a little int here-and-there

Discovered by @phpstan


GrahamCampbell:
Codes aren't necessarily integers actually. PDO exceptions can have strings as codes.

szepeviktor:
I see :(

PR #142, opened 2020-09-17
Use php 7.4 features

snapshotpl:
Since we support php 7.4 as minimum version, we can use php 7.4 features like short function syntax


rosstuck (approved):
YES, YES, SHORTEN ALL OF THE SYNTAX, MWAHAHAHAAHAHAHA

burn

PR #145, opened 2020-10-21
PHP8 Support

flavioheleno:
This PR adds support for PHP8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,17 @@ dist: trusty
 
 php:
     - '7.4'
+    - nightly
 
 matrix:
     include:
         - php: '7.4'
           env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
+        - php: nightly
+          env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
+    fast_finish: true
+    allow_failures:
+        - php: nightly
 
 before_script:
     - travis_retry composer self-update


.travis.yml
```diff
@@ -3,11 +3,17 @@ dist: trusty
...
 php:
     - '7.4'
+    - nightly
 
 matrix:
     include:
         - php: '7.4'
           env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
+        - php: nightly
+          env: 'COMPOSER_FLAGS="--ignore-platform-reqs"'
```
GrahamCampbell, line 13: We should avoid this.

GrahamCampbell (changes requested):

.travis.yml
flavioheleno: Hi @GrahamCampbell I agree, this is just while not all deps support PHP8, I'll remove it from the PR at a later point :-)

GrahamCampbell: Which deps don't support PHP 8?

flavioheleno: Just removed as I checked all deps already support PHP 8 :-D

rosstuck (approved):

.travis.yml
```diff
@@ -3,11 +3,17 @@ dist: trusty
...
 
 matrix:
     include:
         - php: '7.4'
           env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
+        - php: nightly
+          env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
+    fast_finish: true
+    allow_failures:
+        - php: nightly
```
rosstuck, line 16: Nice, didn't know about these options but that's really handy :smile:

rosstuck:
Excellent, thanks once again! :+1:

PR #150, opened 2020-12-09
Support for PHP 8.0 and drop PHP < 7.1

snapshotpl:


tests/Handler/CommandHandlerMiddlewareTest.php
```diff
@@ -10,6 +11,7 @@
 use League\Tactician\Tests\Fixtures\Handler\DynamicMethodsHandler;
 use League\Tactician\Tests\Fixtures\Handler\ConcreteMethodsHandler;
 use Mockery\Adapter\Phpunit\MockeryTestCase;
+use phpDocumentor\Reflection\Types\Void_;
```
GrahamCampbell, line 14: ?

GrahamCampbell (changes requested):

tests/Handler/CommandHandlerMiddlewareTest.php
snapshotpl: Thanks, fixed

tests/Handler/CommandNameExtractor/ClassNameExtractorTest.php
```diff
@@ -12,7 +12,7 @@ class ClassNameExtractorTest extends TestCase
      */
     private $extractor;
 
-    protected function setUp()
+    protected function setUp(): void
```
GrahamCampbell, line 15: Can't do this on PHP 7.0. I think we should just change the PHP version constraint to `^7.1.3 || ^8.0`.

GrahamCampbell (changes requested):

tests/Handler/CommandNameExtractor/ClassNameExtractorTest.php
snapshotpl: Right... Working CI miss there...

rosstuck:
Sorry for the lack of CI, I disabled Travis last night as I'm migrating to GitHub Actions on master. 

I've pushed another branch called `update-ci-for-1.x` that you can rebase on or merge in that should get you working with Github CI.

composer.json
```diff
@@ -10,11 +10,11 @@
         }
     ],
     "require": {
-        "php":  ">=5.5"
+        "php":  "^7.1 || ^8.0"
     },
     "require-dev": {
         "mockery/mockery": "^1.3",
-        "phpunit/phpunit": "^4.8.35",
+        "phpunit/phpunit": "^7.5.20 || ^9.3.8",
```
GrahamCampbell, line 17: Should allow 8.x too.

GrahamCampbell (changes requested):

composer.json
snapshotpl: I don't see any advantage here to allow 8.x. 7.x and 9.x covers all required php versions now.

snapshotpl:
@rosstuck can you look at?

composer.json
```diff
@@ -10,12 +10,12 @@
         }
     ],
     "require": {
-        "php":  ">=5.5"
+        "php":  "^7.1 || ^8.0"
```
acasademont, line 13: Why not just ">=7.1"? These days I'm seeing lots of libraries blocked out of php8 just because of the "^7.x" requirements, even though they would work perfectly fine on php8. 

If php9 came out soon, this library would not install when it would probably work out of the box.

We're using this library on PHP8 without any problems right now thanks to the ">=5.5" requirement, I'd suggest we keep the same ">=" requirement and not "^"

rosstuck:
@snapshotpl Thanks for picking this up, I really appreciate it. It's now tagged as 1.1.0 :)

composer.json
rosstuck: https://github.com/thephpleague/tactician/commit/b8fda250aafcbfdcd4bb213c0397a853452ce3aa Addressed here :)

